PR #103, opened 2026-04-04
Refactor: component audit — adopt admin_page, fix stale link, split oversized components

arsfeld:
## Summary

- **Adopt `admin_page/1` wrapper** in all 8 admin config templates, replacing the duplicated header + tab_nav block (~15 lines each, 8 templates = ~120 lines removed)
- **Fix stale FlareSolverr link** pointing to old `/admin/config?tab=general` route → `/admin/config/settings`
- **Split `admin_indexers_live/components.ex`** (1,252 → 398 LOC) by extracting 3 modal components to `modals.ex` (876 LOC)
- **Split `admin_quality_profiles_live/components.ex`** (1,063 → 187 LOC) by extracting 6 modal components to `modals.ex` (884 LOC)

Part of the Phoenix Architecture Refactor (PR 5, see brainstorm). Independent of PRs #100 and #102.

No behavior changes except the FlareSolverr link fix (corrects a broken link).

## Test plan

- [x] 39 admin tests pass (0 failures, 2 pre-existing skips)
- [x] Full test suite: 3,674 tests, 0 failures
- [x] Compiles with zero errors

No additional operational monitoring required: purely internal code reorganization (template DRY, component file splits). The FlareSolverr link fix corrects an already-broken navigation link.


arsfeld:
Closing this. Audited against master (`0aac12f7`); the three changes bundled here have diverged in three directions.

**Stale FlareSolverr link: already fixed.** `flaresolverr_status_component.ex` no longer exists; it was replaced by `admin_indexers_live/flaresolverr_components.ex`. No `config?tab=` links remain anywhere in `lib/`.

**`admin_page/1` adoption: still worth doing, but re-cut fresh.** `admin_page/1` is on master at `admin_components.ex:100` and exactly one template uses it (`admin_plugins_live`). Nine config templates still duplicate the header + `tab_nav` block, one more than this PR covers, since `path_mappings` was added after April. This is ~10 lines per template on current master, cheaper as a new branch than as a rebase.

**Component splits: the problem persists, but this isn't the right cut.**

- `admin_indexers_live/components.ex` is 1291 LOC now (up from 1252); `admin_quality_profiles_live/components.ex` is 953 (down from 1063).
- The split leaves both new `modals.ex` files at ~880 LOC, still well past the ~500 guideline, so it would need splitting again immediately.
- It cuts by shape rather than sub-domain. `quality_profile_basic_tab` and `quality_profile_standards_tab` are tab content, not modals, and `standards_tab` alone is ~490 lines, so the split relocates the actual problem instead of solving it. `media_live/show/modals.ex` at 1428 LOC shows where that pattern ends up.
- It duplicates `library_health_status_badge_class/1` and `library_health_status_label/1` (all 4 clauses each) into both files, while leaving the pre-existing near-duplicate pairs (`library_indexer_type_badge_class` / `library_indexer_type_badge`, `needs_library_config?` / `needs_library_definition_config?`) untouched.

**Rebase cost.** 992 commits behind and conflicting, with master having edited ~110 lines inside `quality_profile_standards_tab`, a function this PR deletes wholesale and re-adds elsewhere. Git can't rename-detect a partial file split, so that resolution is all by hand.

Broader point on the size work: 30 files in `lib/mydia_web` exceed 500 LOC (`import_media_live/index.ex` leads at 2226), and `.credo.exs` has no module-length check. Hand-splitting two files doesn't move that. If the limit matters it wants a Credo gate with a grandfathered baseline.

Follow-up worth doing on its own: adopt `admin_page/1` across the 9 remaining config templates.

PR #117, opened 2026-05-14
fix(import): handle word-form Season/Episode filenames

arsfeld:
## Summary

- Release parser now extracts episode numbers from filenames like `Show Season 1 Episode 1- Title.mp4` — previously the trailing punctuation caused `parsed.episodes` to come back `nil`.
- `MediaImport.import_file/5` no longer crashes with `FunctionClauseError` in `List.first/2` when `parsed.episodes` is nil; the file is routed to the Issues tab via `match_status: "unresolved_files"`.

## Bug

A user-reported import failure on a TV-show season pack (Off Campus 2026 S01 from PrimeWire). The release-name tokenizer kept the `1-` from `Episode 1- The Deal` as a single token; the verbose-form extractor in `resolver.ex` required a clean integer (`Integer.parse/1 == {n, ""}`) and rejected `"1-"`, leaving `parsed.episodes = nil`. Then `import_file/5` did `List.first(parsed.episodes) || 1` without guarding for nil, so every retry crashed:

    ** (FunctionClauseError) no function clause matching in List.first/2
        # 1: nil
        # 2: nil
        (mydia 0.0.0-dev) lib/mydia/jobs/media_import.ex:942: Mydia.Jobs.MediaImport.import_file/5

After three attempts the Oban job sat in `retryable` state and the download never imported.

## Fix

- `lib/mydia/library/release_parser/resolver.ex`: `verbose_episode_extract/1` matches the leading digit run with a regex so trailing punctuation no longer eats the episode number.
- `lib/mydia/jobs/media_import.ex`: the TV-show branch now requires `parsed.episodes` to be non-nil. A new clause for `season != nil and episodes == nil` returns `{:unresolved, file_info}`, which flows through the existing `flag_unresolved_files/2` path and surfaces the file in the Downloads → Issues tab instead of silently importing as episode 1 or crashing.

## Test plan

- [x] `./dev mix test test/mydia/library/release_parser/` — 361/361 pass
- [x] `./dev mix test test/mydia/library/release_parser_test.exs` — facade test for the full Off Campus filename passes
- [x] `./dev mix test test/mydia/jobs/media_import_test.exs` — 26/26 pass, new regression test verifies match_status: "unresolved_files" and stored parsed metadata
- [ ] Verify on storage: re-trigger MediaImport for the existing stuck Off Campus download (id 2bef0f45) and confirm it either imports (with parser fix) or lands in the Issues tab

### DIFF
--- a/lib/mydia/jobs/media_import.ex
+++ b/lib/mydia/jobs/media_import.ex
@@ -938,7 +938,8 @@ defmodule Mydia.Jobs.MediaImport do
           end
 
         # TV show with parsed episode info - look up the episode
-        {%{type: "tv_show"} = media_item, _, :tv_show, _} when not is_nil(parsed.season) ->
+        {%{type: "tv_show"} = media_item, _, :tv_show, _}
+        when not is_nil(parsed.season) and not is_nil(parsed.episodes) ->
           episode_number = List.first(parsed.episodes) || 1
 
           episode =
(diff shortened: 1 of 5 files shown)


arsfeld:
@copilot apply changes based on the comments in [this thread](https://github.com/getmydia/mydia/pull/117#pullrequestreview-4287057117)

PR #118, opened 2026-05-14
fix(crash-reporting): wire Logger backend and unblock UI toggle

arsfeld:
## Summary

Crash reporting was silently broken on opted-in instances: errors logged via `Logger` never reached the queue, so nothing was ever POSTed to the metadata relay. The admin UI also made the toggle uninteractive whenever `CRASH_REPORTING_ENABLED` was set in the environment, which contradicts how `Mydia.CrashReporter.enabled?/0` actually resolves precedence (DB > env, by design).

## Root cause

`Mydia.CrashReporter.LoggerBackend` was registered via `config :logger, backends: [:console, Mydia.CrashReporter.LoggerBackend]`. In Elixir 1.15+ the legacy `:gen_event` Logger backends were extracted from core into the `:logger_backends` hex package. That dep was missing here, so the `backends:` config was silently ignored — the backend was never installed, and `Mydia.CrashReporter.report/3` (only ever called from that backend) was never invoked. Local `ErrorTracker` still worked because it hooks Logger via telemetry, independently of the legacy registry.

The toggle bug is a separate, smaller defect: `editable={setting.source != :env}` treated env-set values as read-only, but for `crash_reporting.enabled` the runtime gives DB precedence over env so users can opt out at runtime without redeploy.

## Changes

- Add `:logger_backends` dep and install the backend from `Mydia.Application.start/2`. Drop the deprecated `config :logger, backends:` lines in `config/config.exs` and `config/test.exs`.
- Remove the redundant `in_test_env?` guard inside `LoggerBackend` (the `enabled?/0` check already gates everything) and add an integration test that asserts an `:error` log reaches the queue end-to-end.
- Make `Mydia.CrashReporter.stats/0` read the relay URL from the same source as the Sender (`Mydia.Metadata.metadata_relay_url/0`) and surface it in the admin stats panel so operators can verify the target.
- Move the stats panel under the "Crash Reporting" category in the admin UI and only render it when sharing is on.
- For `crash_reporting.enabled`, mark the setting `editable: true` and report `:database` as the source once a DB row exists — so the toggle stays interactive even with `CRASH_REPORTING_ENABLED` set, and the badge truthfully reflects which layer is winning.

## Test plan

- [x] `./dev mix test test/mydia/crash_reporter/ test/mydia_web/live/admin_settings_live_test.exs` — 41 tests, 0 failures.
- [x] `./dev mix compile` — clean.
- [x] `./dev mix format --check-formatted` — clean.
- [ ] Deploy to `storage` podman: confirm `:gen_event.which_handlers(LoggerBackends)` includes `Mydia.CrashReporter.LoggerBackend`.
- [ ] In admin settings UI: toggle "Share Crashes with Developers" off/on with `CRASH_REPORTING_ENABLED` still set in podman; verify the DB row updates and the source badge flips to `DB`.
- [ ] Force a manual report via the admin UI or `Mydia.CrashReporter.submit_manual_report(%RuntimeError{message: "smoke test"})` and confirm it shows up in the can-1 relay's `/errors` dashboard.
- [ ] Verify `METADATA_RELAY_URL` env var on storage points at can-1 (the admin stats panel now displays the resolved URL).


arsfeld:
@copilot apply changes based on the comments in [this thread](https://github.com/getmydia/mydia/pull/118#pullrequestreview-4287099281)

PR #133, opened 2026-05-14
fix(admin-settings): persist boolean toggles as parseable values

arsfeld:
## Summary

The boolean settings toggle in the admin UI was silently broken: clicking it appeared to work, but on refresh the value reverted, and `Mydia.CrashReporter.enabled?/0` kept returning `false` even when the row was clearly present.

Root cause is a collision between LiveView's input-value handling and a bare `<input type="checkbox" phx-click=...>`:

- `lib/mydia_web/live/admin_settings_live/components.ex:148-156` set `phx-value-value={to_string(!@setting.value)}` on the checkbox to convey the desired new state.
- LiveView's `extractMeta` (`deps/phoenix_live_view/assets/js/phoenix_live_view/view.js:1584-1597`) reads `phx-value-*` attributes first and **then** overrides `meta.value` with the input's `el.value`. For a checkbox without an explicit `value` attribute, `el.value` is the HTML default `"on"`.
- So the click sent `params["value"] = "on"`, which the upsert wrote verbatim. `parse_boolean/1` only treats `"true" | "1" | "yes"` as truthy, so the next read returned `false` and the UI flipped back. Production confirmed the same `value: "on"` row on a real instance.

## Fix

- Set `value="true"` on the checkbox so the LiveView override yields a parseable string. Drop the now-redundant `phx-value-value` (LiveView deletes `meta.value` for unchecked checkboxes anyway, so the off click still falls through to the DB-derived toggle path in `index.ex:97-106`).
- Accept `"on"` as truthy in `parse_boolean*` across `Mydia.CrashReporter`, `MydiaWeb.AdminSettingsLive.Index`, and `Mydia.Library.DatabaseHealthCheck` so existing rows written by the broken version self-heal without manual DB intervention.

## Test plan

- [x] `./dev test test/mydia_web/live/admin_settings_live_test.exs` — strengthened the existing crash-reporting toggle test to assert `Mydia.CrashReporter.enabled?/0` returns true after a click.
- [x] Added `self-heals legacy 'on' value written by older versions` regression test that pre-seeds the legacy string and verifies both `CrashReporter.enabled?/0` and that the rendered checkbox is `checked`.
- [x] `./dev test test/mydia/crash_reporter/` — 38 tests pass.
- [x] `./dev test test/mydia/library/database_health_check_test.exs` — 18 tests pass.

### DIFF
--- a/lib/mydia/crash_reporter.ex
+++ b/lib/mydia/crash_reporter.ex
@@ -319,9 +319,12 @@ defmodule Mydia.CrashReporter do
     end
   end
 
+  # `"on"` is accepted because the admin toggle previously persisted the HTML
+  # checkbox default value, and existing config_settings rows may still hold it.
   defp parse_boolean(value) when is_boolean(value), do: value
   defp parse_boolean("true"), do: true
   defp parse_boolean("1"), do: true
   defp parse_boolean("yes"), do: true
+  defp parse_boolean("on"), do: true
   defp parse_boolean(_), do: false
 end
--- a/lib/mydia/library/database_health_check.ex
+++ b/lib/mydia/library/database_health_check.ex
@@ -176,6 +176,7 @@ defmodule Mydia.Library.DatabaseHealthCheck do
   defp parse_boolean_value("true"), do: true
   defp parse_boolean_value("1"), do: true
   defp parse_boolean_value("yes"), do: true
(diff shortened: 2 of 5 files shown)


arsfeld:
@copilot apply changes based on the comments in [this thread](https://github.com/getmydia/mydia/pull/133#pullrequestreview-4292428997)

PR #136, opened 2026-05-14
fix(downloads): self-heal unmatched orphans and stop MediaImport retry loop

arsfeld:
## Summary

Unmatched downloads created by `UntrackedMatcher` (torrents added directly to the download client and discovered by Mydia) have no `media_item_id` and no `library_path_id`. When the torrent later disappears from the client, the row stayed in the DB forever — `MediaImport` retried up to 1000 times with 24h backoff, emitting `:no_library_path` errors for weeks.

In one production instance the activity table shows ~160 such rows from April still failing in May (attempt 42 of 1000), spamming the events table and the queue dedup table with stale rows.

## Why "self-healing" instead of a one-time cleanup

These rows are produced by the regular `UntrackedMatcher` flow whenever the user adds a manual torrent that doesn't auto-match. Cleaning data once doesn't fix the bug — the same condition will recreate stuck rows the next time a manual torrent comes and goes. The fix has to live in the code that creates and processes them.

### `DownloadMonitor.handle_completion/1`
When a completed download is `match_status == "unmatched"`, **skip enqueueing `MediaImport`**. There's no destination, so the import can't do anything useful. The row stays so the user can still manually match it via the Issues tab while the torrent is in the client.

### `DownloadMonitor.perform/1` — new pass
Any unmatched download whose client **confirms** the torrent is gone (`in_client? == false`) is **deleted**, regardless of `completed_at` / `error_message`. Matched downloads still flow through the existing `missing` / `failed` handlers so the user can investigate them in the Issues tab.

### `MediaImport.perform/1`
- If the row no longer exists (`Ecto.NoResultsError`) → return `:ok`. The job is moot; Oban marks it done and the retry loop ends.
- If the row IS there but is an unmatched orphan with no destination → return `{:cancel, :unmatched_no_destination}`. Oban discards the job instead of burning ~1000 retries waiting forever.

### `EnrichedDownload.in_client?`
New field on `EnrichedDownload`. The three enrichment paths in `history.ex` already know whether the torrent is in the client; this field just exposes that fact downstream:

| path                                  | `in_client?` |
| ------------------------------------- | :----------: |
| `enrich_download_with_torrent_status` | `true`       |
| `enrich_download_with_empty_status`   | `false`      |
| `enrich_download_with_unknown_status` | `nil`        |

Without it, the existing `status` field collapsed "client says completed and is still there" with "client says it never had this torrent, but DB has `completed_at`" into the same `"completed"` value — masking exactly the bug we want to catch.

## Safety

- Stale `downloads` rows whose **client is unreachable** stay put (`in_client? == nil`). A network blip cannot trigger mass deletion. Test guards this regression.
- Matched downloads keep their existing missing/failed flow. Only unmatched downloads are auto-deleted.
- The `{:cancel, _}` return for orphan import jobs is the modern Oban discard signal (Oban 2.20+).

## Test plan

- [x] `./dev mix test test/mydia/jobs/download_monitor_test.exs test/mydia/jobs/media_import_test.exs` — 66 tests, 0 failures
- [x] `./dev mix test test/mydia/downloads/ test/mydia/jobs/` — 804 tests, 0 failures
- [x] `mix compile` clean
- [ ] After deploy, confirm activity log no longer shows recurring `Mydia.Jobs.MediaImport failed with {:error, :no_library_path}` events
- [ ] Confirm the existing ~160 unmatched rows drain to zero over the next monitor pass once their client confirms the torrents are gone

### DIFF
--- a/lib/mydia/downloads/history.ex
+++ b/lib/mydia/downloads/history.ex
@@ -258,7 +258,8 @@ defmodule Mydia.Downloads.History do
       import_next_retry_at: download.import_next_retry_at,
       import_failed_at: download.import_failed_at,
       last_progress_at: download.last_progress_at,
-      last_known_bytes: download.last_known_bytes
+      last_known_bytes: download.last_known_bytes,
+      in_client?: true
     })
   end
 
@@ -301,7 +302,9 @@ defmodule Mydia.Downloads.History do
       import_next_retry_at: download.import_next_retry_at,
       import_failed_at: download.import_failed_at,
       last_progress_at: download.last_progress_at,
-      last_known_bytes: download.last_known_bytes
+      last_known_bytes: download.last_known_bytes,
+      # Client unreachable — presence indeterminate.
+      in_client?: nil
     })
   end
 
@@ -355,7 +358,8 @@ defmodule Mydia.Downloads.History do
       import_next_retry_at: download.import_next_retry_at,
       import_failed_at: download.import_failed_at,
       last_progress_at: download.last_progress_at,
-      last_known_bytes: download.last_known_bytes
+      last_known_bytes: download.last_known_bytes,
+      in_client?: false
     })
   end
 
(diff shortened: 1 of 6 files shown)


arsfeld:
@copilot apply changes based on the comments in [this thread](https://github.com/getmydia/mydia/pull/136#pullrequestreview-4292610012)

PR #140, opened 2026-05-15
fix(search): count episodes from DB when metadata episode_count is 0

arsfeld:
## Summary

`should_prefer_season_pack?/3` uses `media_items.metadata.seasons[].episode_count` as the denominator for the 70% threshold that decides between season-pack and individual-episode search. When TVDB scrape leaves that field as `0` (observed in production: **59 of ~93 TV-show seasons** affected), the existing fallback substitutes `missing_count` for the total, yielding 100% missing.

That forces season-pack search for every partial season — even when most episodes are already on disk:

| Show | Season | On disk | Total | % missing | Picked |
|---|---|---|---|---|---|
| Stillwater | 1 | 13 | 24 | 46% | season pack (wrong) |
| Stillwater | 2 | 7 | 13 | 46% | season pack (wrong) |
| Slow Horses | 1 | 5 | 6 | 17% | season pack (wrong) |
| Slow Horses | 5 | 5 | 6 | 17% | season pack (wrong) |

The selected pack is then correctly rejected by `Downloads.Queue.check_for_existing_media_files/3` (some episodes already have media files), but the rejection is logged as `download.failed` at `severity: :error` every cycle of the `all_monitored` job.

**Production impact:** 17,342 lifetime "Download already exists" events, ~250 in the last 36 hours — all the same handful of partial seasons looping.

## Fix

When metadata says `episode_count <= 0`, fall back to counting rows in the `episodes` table — which already has accurate per-season data — instead of treating the broken metadata as authoritative. Partial seasons now correctly route to individual-episode search.

## Test plan

- [x] New unit tests cover the three cases: metadata-zero falls through to DB count, populated metadata is honored, mostly-missing in DB still prefers the pack.
- [x] `./dev mix test test/mydia/jobs/tv_show_search_test.exs` — 34 tests, 0 failures
- [x] `./dev mix format --check-formatted`
- [x] `./dev mix compile`
- [ ] Verify in production after deploy: `download.failed` events with reason `"Download already exists"` should stop accumulating for partial-season shows.

## Out of scope (follow-ups worth considering)

- The misleading `"Download already exists"` reason/message when the actual condition is "media files exist on disk for some episodes in the season" — should probably be `:season_partially_downloaded` at `severity: :info/:warning`, not `:duplicate_download` at `:error`.
- Why TVDB scrape leaves `episode_count: 0` — the real upstream bug. This PR makes the downstream resilient.

### DIFF
--- a/lib/mydia/jobs/tv_show_search.ex
+++ b/lib/mydia/jobs/tv_show_search.ex
@@ -656,7 +656,9 @@ defmodule Mydia.Jobs.TVShowSearch do
     final_count
   end
 
-  defp should_prefer_season_pack?(missing_episodes, media_item, season_number) do
+  @doc false
+  # Public for unit testing — internal helper.
+  def should_prefer_season_pack?(missing_episodes, media_item, season_number) do
     missing_count = length(missing_episodes)
 
     # Try to get total episode count from metadata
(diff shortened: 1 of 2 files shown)


arsfeld:
@copilot apply changes based on the comments in [this thread](https://github.com/getmydia/mydia/pull/140#pullrequestreview-4299031779)

PR #151, opened 2026-05-17
fix: stop retry loop on unimportable malware downloads

arsfeld:
## Summary

Two compounding bugs let malware torrents named like `From.S04E05.1080p.WEB.h264-ETHEL.exe` sit in the import retry queue for days. With `max_attempts: 1000` and 24-hour backoff past attempt 7, a single bad grab can rack up double-digit retries before anyone notices. Spotted in production with four such rows on one host, the worst at `retry_count: 11`.

### Two root causes

1. **`ReleaseValidator` didn't recognize executable extensions as a malware tell.** Torrents whose name ended in `.exe` parsed cleanly and entered the queue.
2. **`:no_importable_files` was treated as a transient failure.** The importer correctly rejected the `.exe` payload every time, but the retry loop never gave up.

### Changes

- **`ReleaseValidator`** rejects release names ending in `.exe .scr .bat .cmd .com .msi .vbs .js .jar .pif` as `:suspicious_extension`. A legitimate video release never carries one of these, and several public indexers periodically post malware under real-looking release-group names (e.g. `ETHEL`) to slip past automatic-grab filters.
- **`MediaImport`** gains a `terminal_failure?/2` classifier. When a failure is structurally unfixable, `perform/1` returns `{:cancel, reason}` (same escape hatch already used for unmatched orphans) and clears `import_next_retry_at` so the UI shows "gave up" instead of advertising a retry that will not fire:
  - `:no_importable_files` is terminal on the **first** attempt. The error only fires after a *completed* torrent has been scanned and zero files matched the video-extension whitelist. Re-scanning the same finished torrent will deterministically produce the same result.
  - `{:path_not_found, _}` and `{:path_not_accessible, _}` are terminal after **3** attempts so a transient mount blip still gets a chance to recover.
  - Every other failure keeps current backoff behavior.

### Behavior on already-stuck rows

No data migration needed. On each affected row's next scheduled retry, MediaImport will hit the same error, classify it terminal, return `{:cancel, ...}`, and Oban will discard the job. The download row stays (with `import_next_retry_at = nil`) so it surfaces in the Issues tab as a permanent failure to be cleared manually, instead of quietly retrying for weeks.

## Test plan

- [x] `mix test test/mydia/downloads/release_validator_test.exs` (48 tests, 9 new for `.exe` and friends)
- [x] `mix test test/mydia/jobs/media_import_test.exs` (33 tests, 1 new end-to-end terminal-failure case that drops a `.exe` in a save_path dir and asserts `{:cancel, :no_importable_files}` plus `import_next_retry_at: nil`)
- [x] Full `mix precommit` (4506 tests, 0 failures)

### DIFF
--- a/lib/mydia/downloads/release_validator.ex
+++ b/lib/mydia/downloads/release_validator.ex
@@ -109,7 +109,12 @@ defmodule Mydia.Downloads.ReleaseValidator do
   @suspicious_extensions ~w(.exe .scr .bat .cmd .com .msi .vbs .js .jar .pif)
 
   defp suspicious_extension?(name) do
-    ext = name |> Path.extname() |> String.downcase()
+    ext =
+      name
+      |> String.trim()
+      |> Path.extname()
+      |> String.downcase()
+
     ext in @suspicious_extensions
   end
 
--- a/lib/mydia/downloads/release_validator.ex
+++ b/lib/mydia/downloads/release_validator.ex
@@ -28,6 +28,12 @@ defmodule Mydia.Downloads.ReleaseValidator do
      - Example: `yenc movie title`
      - These are raw usenet posts, not proper releases
 
+  6. **Suspicious executable extensions** - Release name ends in an executable
+     extension (.exe, .scr, .bat, .cmd, .com, .msi, .vbs, .js, .jar, .pif)
+     - Example: `From.S04E05.1080p.WEB.h264-ETHEL.exe`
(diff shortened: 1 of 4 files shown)


arsfeld:
@copilot apply changes based on the comments in [this thread](https://github.com/getmydia/mydia/pull/151#pullrequestreview-4304740747)

PR #196, opened 2026-06-08
fix(metadata): respect configured language for TV series, seasons, and episodes

### LINKED ISSUE #194: Metadata Language
Movies takes correctly the defined metadata language, but tvseries always take english.

arsfeld:
## Problem

Fixes #194 — movies correctly use the configured metadata language, but TV series, seasons, and episodes always rendered in English.

## Root cause

Movies and TV use different providers and select language differently:

- **Movies → TMDB:** language selected server-side (`language=` query param). Worked.
- **TV → TVDB:** TVDB returns *all* translations in one bundle (`meta=translations`) and the client picks one. That selection was hardcoded to English (`language == "eng"`) in three places, and the configured language was never threaded into the TVDB fetch path (the `opts` arg was ignored).

## Changes

- **`Mydia.Metadata.LanguageCode`** (new): maps the configured BCP-47/ISO 639-1 value (`en-US`, `de`) to TVDB's ISO 639-2/T 3-letter codes (`eng`, `deu`), and provides a shared, ordered translation selector — replacing three duplicated hardcoded-English extractors.
- Threads the configured language into the TVDB series, season, and episode fetch paths.
- Selection follows the fallback chain **configured language → original language → English → raw field**.
- **Cache correctness:** the metadata cache (`fetch_by_id_cached`, `fetch_season_cached`) and the refresh-path cache invalidation now key by the configured language instead of defaulting to `en-US`, so non-English libraries no longer read English-cached entries and mixed-language libraries don't collide.
- Original language is threaded through all three season-fetch paths (enricher, `Media.create_episodes_for_season`, `ProviderSwitch`) for consistent fallback.

The TMDB/movie path, the language-setting UI, and subtitle language are untouched.

## Testing

- New unit tests for `LanguageCode` (mapping, normalization, selection), the season/episode struct selection, and cache-key-by-language behavior.
- An end-to-end integration test: a Korean-configured show whose TVDB record has only Japanese + English translations correctly falls back to **Japanese** (the original language), proving the threading works across the full enrich path.
- `mix format`, `credo --strict`, and `compile --warnings-as-errors` all clean.

No special monitoring required. This is read-path metadata selection. The visible signal: non-English libraries see localized TV titles/overviews after their next metadata refresh. Entries cached under `en-US` before this change age out at their 24h TTL (no forced backfill); if a library should update sooner, a manual refresh re-fetches under the correct language.

### DIFF
--- a/lib/mydia/library/metadata_enricher.ex
+++ b/lib/mydia/library/metadata_enricher.ex
@@ -396,6 +396,11 @@ defmodule Mydia.Library.MetadataEnricher do
     # Get seasons list from metadata (includes tvdb_season_id if from TVDB)
     seasons = get_seasons_list(media_item.metadata)
 
+    # The show's original language lets the TVDB season/episode fetch prefer the
+    # original-language translation before falling back to English (TVDB stores
+    # it as a 3-letter code, matching the translation bundle's keys).
+    original_language = metadata_original_language(media_item.metadata)
+
     if seasons != [] do
       # Fetch and create/update all episodes
       Enum.each(seasons, fn season ->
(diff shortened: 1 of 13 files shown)


arsfeld:
Addressed both Copilot comments in 0b258f5c:
- `original_language_from/1` now also matches string-key metadata maps, so the original-language fallback tier isn't skipped on paths where `media_item.metadata` is a string-key map.
- `search_cached/3` now defaults its cache-key language from the provider config instead of a literal `en-US`, closing the same cache-collision gap.

PR #197, opened 2026-06-08
fix(metadata): match both por and pt TVDB translation codes for Portuguese

arsfeld:
## Problem

Follow-up to #196, found by validating against real TVDB data on a live instance.

The merged fix mapped each configured language to a single TVDB (ISO 639-2/T) 3-letter code. But TVDB's translation keys are **inconsistent**: while almost every language uses the 3-letter code, **Portuguese appears as both `por` and `pt`**, and some shows carry only one. For example, *Frieren: Beyond Journey's End* (tvdb `424536`) has only `pt` in its translation bundle — so a Portuguese-configured library requesting `por` matched nothing and fell back to English.

## Validation

Inspected the real `nameTranslations` codes across 5 popular shows on a live instance. Distinct codes:

```
ara cat ces dan deu ell eng fas fin fra heb hin hrv hun ind ita jpn kan kat
kor mal msa nld nor pol por pt ron rus spa srp swe tam tel tgl tha tur ukr
vie zho zhtw
```

`pt` is the **only** 2-letter exception — every other supported language (`es→spa`, `fr→fra`, `de→deu`, `it→ita`, `ja→jpn`, `zh→zho`, `ko→kor`, `ru→rus`, `en→eng`) is confirmed 3-letter. The `deu`/`ger` and `zho`/`chi` variants I'd flagged as risks resolve to `deu` and `zho` — matching the existing mapping.

## Change

Replace the single-code mapping (`to_tvdb_code/1` in the selection path) with `tvdb_candidates/1`, which expands each configured/original language to **both** its 3-letter and 2-letter forms (`pt-BR → ["por", "pt"]`). The form TVDB didn't use simply never matches, so this is harmless for the other nine languages and robust to TVDB's inconsistency going forward.

Verified end-to-end against Frieren's real bundle: a `pt-BR` library now selects `"Frieren e a Jornada para o Além"` instead of the English title.

No special monitoring required (read-path metadata selection). Portuguese libraries will see Portuguese TV titles/overviews after their next refresh, including on `pt`-only shows.

### DIFF
--- a/lib/mydia/metadata/language_code.ex
+++ b/lib/mydia/metadata/language_code.ex
@@ -55,37 +55,35 @@ defmodule Mydia.Metadata.LanguageCode do
   def to_tvdb_code(_), do: nil
 
   @doc """
-  Normalizes a language code that may be either a 2-letter (ISO 639-1, possibly
-  region-suffixed) or an already-3-letter (ISO 639-2/T) value into the TVDB
-  3-letter form. Use for values whose width depends on their source — e.g. a
-  show's `original_language`, which TVDB stores as `"jpn"` but TMDB stores as
-  `"ja"`. Returns `nil` for blank or unrecognized input.
+  Returns the ordered list of TVDB language codes to try for a configured or
+  original-language value. TVDB's translation keys are mostly ISO 639-2/T
+  3-letter codes, but it is inconsistent — Portuguese appears as both `"por"`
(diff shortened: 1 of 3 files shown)


manueljortega:
using :latest still not working for spanish. Tried both with "es" and "es-ES". Movies scraped correctly but Series keeps in english.
Captura de pantalla 2026-06-09 080834
Captura de pantalla 2026-06-09 080722

PR #221, opened 2026-06-25
Fix TorBox torrent upload handling

ppoloskov:
## Summary

Fixes two TorBox provider edge cases found while investigating intermittent Mydia download failures:

- use Req-compatible multipart syntax for `.torrent` file uploads so real torrent files do not crash in `Req.Utils.encode_form_part/2`
- keep TorBox no-seed/checking/stalled torrent states active instead of treating them as immediate terminal client failures
- add Bypass regression coverage for torrent-file multipart upload, no-seed stall handling, and `missingFiles` remaining terminal
- document the TorBox/Req gotchas in README, AGENTS, and handover notes

## Why

Live logs showed `.torrent` uploads through the TorBox adapter crashing before submission with:

```text
FunctionClauseError no function clause matching in Req.Utils.encode_form_part/2
```

The adapter was building file multipart fields as a three-element tuple. Req expects file fields in the `{body, opts}` shape, for example:

```elixir
file: {bin, filename: "release.torrent", content_type: "application/x-bittorrent"}
```

The same investigation also found that TorBox can accept one release while another stalls or fails provider-side. TorBox official docs describe `stalled (no seeds)` as a torrent still trying to download with no connected seeds, not as an immediate hard failure. Mapping that state to `:downloading` lets the existing Mydia monitor and stall detector observe it over time before escalating.

## Official Docs Checked

Confirmed against TorBox official docs:

- `https://api-docs.torbox.app/` / published Postman collection
- `https://api.torbox.app/openapi.json`

Relevant doc points:

- `POST /torrents/createtorrent` accepts either `magnet` or multipart `file`
- torrent `download_state: "stalled (no seeds)"` means the torrent is attempting to download but currently has no connected seeds
- `missingFiles` remains a terminal/error state in this patch

## Verification

Passed:

```text
mix format --check-formatted lib/mydia/downloads/client/debrid/providers/tor_box.ex test/mydia/downloads/client/debrid/providers/tor_box_test.exs
mix test test/mydia/downloads/client/debrid/providers/tor_box_test.exs
# 14 tests, 0 failures
mix test test/mydia/downloads/client/debrid/shared_test.exs test/mydia/downloads/client/debrid_test.exs
# 56 tests, 0 failures
git diff --check
mix compile
mix ecto.migrate
mix phx.server + HEAD http://127.0.0.1:4400/ -> 302 /auth/login
```

Notes:

- verified via direct Nix shell because local `./dev`/devenv evaluation is currently blocked by a Nix restricted `system` setting/trust issue
- backend app compiles and serves HTTP after migrations
- dev frontend watchers still need the normal dev environment setup: `assets/node_modules`, Flutter on PATH, and optional `wasm32-wasip2` for bundled plugin WASM artifacts


arsfeld:
Hey @ppoloskov, thanks for the PR. A couple of things before I merge:

1. **`handover.md`** — could you drop this from the PR? It's a point-in-time investigation log tied to your own setup (k8s/PVC notes, a specific blacklisted release), so it doesn't really belong committed to the repo root. The AGENTS.md "Download client gotchas" section already captures the durable lessons, which is the right home for them.

2. **`"checking"` → `:downloading`** — this one isn't mentioned in the PR description alongside the documented stalled states, and it's a behavior change: previously `"checking"` hit the catch-all `:queued` (not stall-tracked), and now a hash-checking torrent makes no byte progress while counting as `"downloading"`, so in theory it could escalate as a stall. In practice checking is brief so it's basically harmless, but I want it to be intentional rather than incidental.

3. **Minor, not blocking:** RealDebrid still maps `"dead"` (no seeders) straight to `:error` in `real_debrid.ex`, which is the same immediate-blacklist pattern you're fixing here for TorBox. The semantics might genuinely differ (RD may not retry a dead torrent the way TorBox does), so I'm not asking you to touch it — just noting it so we keep the no-seed handling consistent across providers later.

One thing to keep in mind: I don't have a TorBox account, so I can't actually test this myself — that's why it's marked bypass-only in the code. I'm glad someone's using it and putting it through its paces, but it does mean I'm relying on you to confirm it works against a live account. So if you can verify the `.torrent` upload and the stall behavior on your end, that'd be a big help.

Once `handover.md` is out I'm happy to merge.

arsfeld:
Thanks for tracking this down, the root cause diagnosis is exactly right. I reviewed it and pushed one commit onto the branch (20b0f387) rather than leaving review notes, since most of it was mechanical.

**The multipart bug wasn't TorBox-specific.** `Req.Utils.encode_form_part/2` only has clauses for `{name, {value, opts}}` and `{name, value}`, so the three-element tuple fails on *every* file upload, not just real `.torrent` files. Premiumize (`:src`) and AllDebrid (`:"files[]"`) built their fields the same way and had no file-upload coverage, only magnet tests. Both are fixed, and all three providers now have a Bypass regression test. I confirmed both new sibling tests reproduce the identical `FunctionClauseError` before the fix:

```
1) test submit_torrent/2 torrent file upload uses Req-compatible multipart body (PremiumizeTest)
   ** (FunctionClauseError) no function clause matching in Req.Utils.encode_form_part/2
2) test submit_torrent/2 torrent file upload uses Req-compatible multipart body (AllDebridTest)
   ** (FunctionClauseError) no function clause matching in Req.Utils.encode_form_part/2
```

Real-Debrid is unaffected (raw PUT body), as are qBittorrent and SABnzbd, which already used the correct shape.

Smaller changes in the same commit:

- The fixtures now carry raw invalid-UTF8 bytes, mirroring the SHA-1 `pieces` field of a real `.torrent`, so a body mangled by an encoding step would fail too.
- Added coverage for `stalledDL` and `checking`, which the state remap introduced without tests. Folded the state assertions into the table-driven idiom the AllDebrid tests already use.
- Moved the debrid caveat from the README into `docs/user-guide/download-clients.md`. "Bypass-only" refers to the test library, which doesn't mean much to an operator deciding whether to enable TorBox.
- Dropped `handover.md`. The durable parts are already in the AGENTS.md gotchas section, and I filed the `raw_status` follow-up as #237 so it doesn't get lost.

**One thing I deliberately left alone, worth your call.** Debrid clients default to a 1440-minute grace window with a 3x escalation multiplier, so a torrent TorBox already reports as having zero seeds now holds its episode slot for roughly four days before failing and freeing the release for an alternative grab. Failing on the first poll was clearly too aggressive, but four days may be too passive in the other direction. I didn't want to change the tuning unilaterally. Happy to leave it as-is if you think the stall detector handles it well enough in practice.

Local verification: `mix test test/mydia/downloads/client/debrid/` is 117 tests, 0 failures, and `mix format --check-formatted` is clean. Note that `test/mydia_web/live/downloads_live/path_mapping_issue_test.exs` times out on four tests, but it does so identically on your branch before my commit, so it's pre-existing and unrelated.

PR #224, opened 2026-06-29
feat(media): add progress filter to media library pages

aescolastico:
Proposal adds a new Media page progress filter for movies and shows, allowing admins to quickly find missing, partially downloaded, or fully downloaded items in large libraries. The filter follows the existing LiveView pattern that the others use; applied server-side to the loaded media items rather than persisted to the database. So no db migrations are needed.

The available dropdown options adhere the existing semantics present in the UI: All, Missing, Partial, Downloaded

2 added tests cover the filtering behavior for both media types.

Screenshot 2026-06-28 at 9 45 55 PM

Screenshot 2026-06-28 at 9 26 01 PM


aescolastico:
hi @arsfeld,
any interest in this feature? should be a pretty standard extension of the existing norm.

arsfeld:
Thanks for this! Sorry for the slow reply, this one sat longer than it should have.

Really nicely done on the plumbing. You filtered before pagination and remembered to update select_all and ctrl+A too, which is exactly the bit I'd have expected to get missed.

I pushed one change on top before merging. The filter had its own status logic, which meant it could disagree with the status badge sitting right next to each item: a movie that was actively downloading showed up under Missing (so you'd go grab something already in flight), unmonitored stuff got counted as downloaded or missing, and shows with unaired episodes could read Downloaded while the badge said Partial. Swapped it over to Media.get_media_status/1, the same function the badge already uses in that LiveView, so the two can't drift apart. Added a couple of tests for those cases, and it ended up deleting more than it added.

Also merged master in since the branch had fallen a fair bit behind. Everything's green, merging now. Thanks again!

aescolastico:
> The filter had its own status logic, which meant it could disagree with the status badge sitting right next to each item

Oh nice, I didn't notice there was a `get_media_status` that handled that already. I'll make sure to familiarize myself more with the existing code structures for the next one. Thanks for fixing, responding and merging!

PR #225, opened 2026-06-29
fix(ui): restore browser favicon

aescolastico:
Adds explicit favicon, PNG icon, mask icon, and Apple touch icon metadata so browsers can discover the app icon. The PWA manifest alone was not enough for any of the browsers I tested. 
Spot tested this change on a few browsers and devices: Brave(macOS), and Safari(macOS), and Edge(Windows). I don't use chrome, but it should resolve it there as well. 
image


aescolastico:
@arsfeld more false positive test failures?

arsfeld:
hey @aescolastico, thanks for this! the red CI was all unrelated flaky tests, nothing to do with your change. merged it in. thanks again!

PR #230, opened 2026-07-28
fix(downloads): stop rqbit shared-dir series contamination

arsfeld:
## Summary

Stops a download client's shared output directory from contaminating series libraries, and fixes the three amplifiers that turned one such misplacement into ~980 GiB of duplicated data on a production instance.

- **Scope every client to its own files.** rqbit populates `DownloadStatus.files` from the torrent's file list; qBittorrent from `content_path`; rTorrent from `d.base_path`. Transmission, nzbget and sabnzbd already report per-download paths.
- **Close the ways back into the recursive shared-root walk.** An empty scoped file list no longer falls back to it, and the fallback refuses a `save_path` equal to the client's configured download root.
- **Stop retries duplicating files.** Deterministic conflict filenames, `:expected_size` passed to `place_file`, and a terminal cap on `:partial_import`.

### The scoping bug

`librqbit`'s `get_default_subfolder_for_torrent` returns `None` when `files.len() < 2`, so single-file torrents inherit the shared session root:

```json
{"name":"Silo.S03E01.1080p.x265-ELiTE", "output_folder":".../rqbit/Silo.S03E01.1080p.x265-ELiTE"}
{"name":"Silo.S03E02....NeoNoir.mkv",   "output_folder":".../rqbit/"}
```

`MediaImport` then recursively listed that root and copied every neighbour into whichever show was importing. On one instance: 913 stray files, ~980 GiB, 7 shows plus a movie folder, 7 independent copies of a single episode. The copies were real, not reflinks (`Exclusive 2.55GiB / Set shared 0.00B`, 7 inodes, all `links=1`).

qBittorrent had the identical bug: `save_path` is the *containing* directory, so a single-file torrent reports the download root exactly as rqbit did. `content_path` is already returned by `/torrents/info`, so scoping costs no extra request. rTorrent's `d.directory` has the same shape; `d.base_path` is the torrent's own data.

### The amplifiers

Misplacement alone moves files. These made it a terabyte:

- `generate_unique_path/1` appended a unix timestamp. With `rename_files` on, every foreign file parsed as `S01E01` renamed to the same target, sizes differed, and each of up to 1000 retries minted a brand new multi-GB copy. Conflicts now take the lowest free `name.N.ext` and **adopt an existing same-size sibling**, so retries converge. Bounded at 10 suffixes, then terminal.
- `place_file` was called without `:expected_size`, so the already-present short-circuit could never fire — `File.ln` hit EEXIST and fell through to `File.cp`. Both call sites now pass it.
- `:partial_import` had no `terminal_failure?/2` clause under `max_attempts: 1000`. Terminal after 3 attempts.

### What was dropped

**The foreign-title guard.** It cannot fire on title-less releases, which is how the bug recurred mid-cleanup — `Mushoku Tensei S01 .../S01E01-Jobless Reincarnation.mkv` has no series name to compare, and three such files landed in *House of the Dragon* next to the real episodes. This naming style is not rare (Frieren, Solo Leveling, Halt and Catch Fire). The 0.5 threshold is also the parser's global `suggest_threshold` and is noisy on short titles:

| Pair | Score | Caught |
|---|---|---|
| `Star Trek The Original Series` vs `Silo` | 0.523 | no |
| `Andor` vs `Ahsoka` | 0.578 | no |
| `Foundation` vs `Fallout` | 0.638 | no |
| `The Last of Us` vs `The Lost Boys` | 0.728 | no |

And a false positive skipping every candidate returned `:no_importable_files`, which is terminal — trading silent contamination for silent permanent import failure. Scoping to the real file list is the load-bearing fix.

**`MisplacedSeriesHealer`.** It did no database work at all: no `Repo`, no `MediaFile`, no `Mydia.Library`. Every file it moved left a `media_files` row pointing at a path that no longer existed, with strays still bound to the wrong show's episodes — 1233 mis-bound rows on this incident. It also created `_misplaced` inside the library root with no `scanner.ex` exclusion, so the next scan re-ingested the quarantine, and on `auto_organize` libraries it compared filenames against the literal string `"Anime"`. The cleanup was done directly instead.

## New error states

| Reason | Terminal | Surfaced as |
|---|---|---|
| `{:scoped_files_missing, count}` | after 3 attempts | client listed N files, none exist locally — add a path mapping |
| `{:save_path_is_download_root, path}` | immediately | client reports its own download root — enable per-torrent subfolders |
| `{:too_many_conflicts, path}` | immediately | too many files claiming one filename — check the download's contents |

## Test plan

- [x] `mix precommit` — 5243 tests, only 4 pre-existing `path_mapping_issue_test.exs` timeouts, verified failing identically on the unmodified branch
- [x] New `media_import_scoping_test.exs` drives the job through a stub adapter, so `resolve_status_files/2`'s `files` branch and `file_entry_if_exists/1` are actually exercised — the previous test registered `host: "nonexistent.invalid"`, so `get_status` failed and it ran the `save_path` fallback instead of the new code
- [x] Adapter coverage for qBittorrent `content_path` (single-file, multi-file, pre-2.6.1, `content_path == save_path`) and rTorrent `d.base_path`
- [x] Regression test that a re-run adopts the earlier conflict copy instead of duplicating it
- [ ] Deploy to galactica and confirm a new Silo import only moves Silo files


arsfeld:
This bug fired on a production instance and I just finished cleaning up after it. Posting the incident data and a review of this PR against it, since the real-world shape turned up a couple of cases the current approach does not cover.

## What the incident actually did

`MediaImport` swept the shared rqbit root and copied every neighbour file into whichever show was importing. Damage on one instance:

| | |
|---|---|
| Stray files created | 913 |
| Duplicated data | ~980 GiB |
| Shows contaminated | 7, plus one movie folder |
| Copies of a single episode | 7, across 7 different show folders |

The copies were real and independent, not hardlinks and not btrfs reflinks: `btrfs fi du` reported `Exclusive 2.55GiB / Set shared 0.00B` per copy, 7 distinct inodes, all `links=1`. `Movies/Disclosure Day (2026)/` ended up holding 213 files of which exactly one was the actual movie. The entire Star Trek TOS series landed inside the `Silo` folder, bound to Silo's episode rows.

## The diagnosis in this PR is right

Confirmed against the live API. Both shapes side by side, same client, same session:

```json
{"name":"Silo.S03E01.1080p.x265-ELiTE",  "output_folder":".../rqbit/Silo.S03E01.1080p.x265-ELiTE"}
{"name":"Silo.S03E02....NeoNoir.mkv",    "output_folder":".../rqbit/"}
```

And upstream in `librqbit/src/session.rs`, `get_default_subfolder_for_torrent`:

```rust
if files.len() < 2 { return Ok(None) }
```

So single-file torrents genuinely inherit the shared session root. The `files:` plumbing in `rqbit.ex` plus `resolve_status_files/2` closes that correctly, and the `Path.join(output_folder, name)` is right for multi-file torrents too, since rqbit's `name` is the full relative path rather than a basename.

## What is still open

**1. qBittorrent has the identical bug and is untouched.** `qbittorrent.ex:458` uses `save_path`, which is the containing directory, so a single-file torrent reports the download root exactly as rqbit did. `DownloadStatus.files` stays `nil` and the importer falls back to the recursive walk. `content_path` and `/api/v2/torrents/files` are both available and unused. Worth folding in here, it is a small change and qBittorrent is the more common client. `rtorrent.ex:338` uses raw `d.directory` and should be checked too. Transmission already joins `downloadDir` + `name` and looks fine.

**2. The recursive fallback at `media_import.ex:283` still sweeps the shared root**, and `args.save_path` comes straight from the client-reported path via `download_monitor.ex:725`. Any transient client error re-enters the original failure mode. The new `{:error, :no_files}` at `media_import.ex:569` is also a fresh way into it. Refusing a `save_path` equal to the client's configured download root would close this cheaply.

**3. The title guard cannot fire on title-less releases.** This is the one I would most want addressed, because it showed up live. The bug recurred on the instance mid-cleanup, and one of the new torrents was:

```
Mushoku Tensei S01 1080p Dual Audio BD remux FLAC-TTGA
  S01E01-Jobless Reincarnation.mkv
  S01E02-Teacher.mkv
  S01E03-A Friend.mkv
```

There is no series name in those filenames at all, so `foreign_title_mismatch?/1` has nothing to compare and the guard is a no-op. Three of them landed in `House of the Dragon/Season 01` next to the real episodes, separable only by file size. This naming style is not rare: Frieren, Solo Leveling and Halt and Catch Fire all use it in the same library.

**4. The 0.5 jaro threshold is noisy on short titles.** Running `Text.title_similarity/2` on realistic pairs:

| Pair | Score | Caught |
|---|---|---|
| `Star Trek TOS` vs `Silo` | 0.442 | yes |
| `Star Trek The Original Series` vs `Silo` | 0.523 | no |
| `Andor` vs `Ahsoka` | 0.578 | no |
| `Foundation` vs `Fallout` | 0.638 | no |
| `The Last of Us` vs `The Lost Boys` | 0.728 | no |

Same show, different release naming, different outcome.

**5. A false positive is now a permanent failure.** If the guard skips every candidate, `media_import.ex:800-806` returns `{:error, :no_importable_files}`, which `terminal_failure?/2` treats as terminal, so the job cancels with no retry. That trades silent contamination for silent permanent import failure. Lowering the threshold or gating on an explicit `original_title` / alt-title check would avoid that.

## The amplifiers are what made this a terabyte

The scoping bug alone misplaces files. These three are what turned it into ~980 GiB, and none are addressed here:

- `max_attempts: 1000` with no `:partial_import` clause in `terminal_failure?/2`, so every failure retries and re-copies.
- `place_file` is called without `:expected_size` at `media_import.ex:1452`, so the already-present short-circuit at `file_organizer.ex:285` can never fire. On retry `File.ln` hits EEXIST and falls through to `File.cp`.
- `handle_file_conflict/6` calls `generate_unique_path/1`, which appends a unix timestamp. With `rename_files` on, every foreign file parsed as `S01E01` renames to the same target, sizes differ, and each retry mints a brand new multi-GB copy.

## Test coverage

`media_import_test.exs:1417` registers a client at `host: "nonexistent.invalid"`, so `get_status` fails and the test runs through the `save_path` fallback rather than the new code. `resolve_status_files/2`'s `files` branch and `file_entry_if_exists/1` have no coverage there. A test that stubs a `DownloadStatus` with `files:` set and `save_path` pointing at a shared root containing a neighbouring show's file would pin the actual fix. CI also has not run on the branch yet.

## On the healer

I would land `MisplacedSeriesHealer` separately. It does no database work at all: no `Repo`, no `MediaFile`, no `Mydia.Library`. It moves files and leaves every `media_files` row pointing at a path that no longer exists, with strays still bound to the wrong show's episodes. On this incident that would have left 1233 mis-bound rows behind after the "repair", so I ended up doing the cleanup directly instead. Two other things to sort before it ships: `_misplaced` is created inside the library root and `scanner.ex` has no exclusion for it, so the next scan re-ingests the quarantine; and on `auto_organize` libraries it treats the category directory as a show folder and compares filenames against the literal string `"Anime"`.

## Suggestion

Ship the rqbit scoping plus the qBittorrent `content_path` fix and the end-to-end test as one unit, since that is the part that is clearly correct. Then the three amplifiers. The title guard needs a decision on threshold and failure mode first, and it will not help title-less releases either way, so scoping the import to the real file list stays the load-bearing fix.

arsfeld:
All five open items plus the three amplifiers are addressed in 96d75b00; the healer is removed in 61f01011. Point by point:

**1. qBittorrent / rTorrent.** Both fixed. qBittorrent uses `content_path` rather than `/api/v2/torrents/files` — it is already in the `/torrents/info` response, so scoping costs no extra request, and it resolves to the torrent's own file (single-file) or its own root folder (multi-file), which `file_entry_if_exists/1` then recurses. Servers predating Web API 2.6.1, and multi-file torrents added with subfolder creation disabled where `content_path == save_path`, fall back as before. rTorrent now requests `d.base_path` alongside `d.directory`. Confirmed Transmission is fine; nzbget (`dest_dir`) and sabnzbd (`storage`) also report per-job directories.

**2. The recursive fallback.** Both entrances closed. The empty-scoped-list case is now `{:scoped_files_missing, count}` and refuses the fallback outright rather than falling into it — the client has already told us the directory is not scoped to this download. And the fallback refuses a `save_path` equal to the client's configured `download_directory`, compared both raw and after path mapping.

**3, 4, 5. The title guard.** Dropped. Your title-less case is the deciding one: the guard cannot fire on it either way, and it recurred live. Adding to that, the 0.5 threshold is `Config.suggest_threshold()` — the parser's global — so retuning it for this guard would move `binding_suspect` everywhere, and the terminal `:no_importable_files` on a false positive trades silent contamination for silent permanent failure. Scoping to the real file list is doing the work.

**The amplifiers.** All three:

- `generate_unique_path/1`'s timestamp is gone. Conflicts take the lowest free `name.N.ext` and adopt an existing same-size sibling, so a retry finds what the previous attempt wrote instead of minting another copy. Bounded at 10 suffixes, then `{:too_many_conflicts, path}`, terminal.
- `:expected_size` now passed at both `place_file` call sites.
- `:partial_import` terminal after 3 attempts.

**Test coverage.** You were right that `media_import_test.exs:1417` never reached the new code. New `media_import_scoping_test.exs` registers a stub adapter so `resolve_status_files/2`'s `files` branch and `file_entry_if_exists/1` actually run: a shared root holding a neighbour's file, a multi-file torrent's own directory, missing scoped files, and both fallback outcomes. Plus adapter-level coverage for `content_path` and `d.base_path`, and a regression test that a re-run adopts the earlier conflict copy rather than duplicating it.

**The healer.** Deleted rather than split out — the dangling-`media_files` problem, the `_misplaced` scanner re-ingest, and the `auto_organize` category-dir bug are all real, and with the cleanup already done by hand the remaining value did not justify a title-similarity file mover on a live library.

`mix precommit`: 5243 tests, 4 failures, all in `path_mapping_issue_test.exs` and all `ExUnit.TimeoutError`. I stashed the branch back to 01d7b69a and confirmed those four fail identically there, so they are pre-existing and not from this change.

Still open from your list: the galactica deploy verification.

arsfeld:
Re-reviewed at `df1ed10a` against the production incident. This closes 9 of the issues I raised directly (qBittorrent `content_path`, rtorrent `d.base_path`, the shared-root guard, `{:scoped_files_missing, _}` no longer falling through, the scoping test, `:partial_import` termination, `:expected_size`, and the conflict-suffix cap) and resolves 8 more by deletion.

Dropping the title-similarity guard was the right call, not just a scope cut. It was a heuristic patch for a scoping problem, and it could not have worked: the release that re-triggered this on my instance was `Mushoku Tensei S01 1080p Dual Audio BD remux FLAC-TTGA`, whose files are named `S01E01-Jobless Reincarnation.mkv`, `S01E02-Teacher.mkv`. There is no series name in the filename to compare, so the guard was a no-op for exactly the case that hurt. Frieren, Solo Leveling and Halt and Catch Fire all use that naming style too. Asserting scope instead of guessing identity is the correct fix.

One change I would want before merge.

## The new guard is config-dependent and silently inert when unset

`shared_download_root?/2` is the main safety net, but it only works if `download_directory` is populated:

```elixir
defp shared_download_root?(save_path, %{download_directory: root})
     when is_binary(save_path) and save_path != "" and is_binary(root) and root != "" do
  same_dir?(save_path, root) or
    same_dir?(mapped_path_quiet(save_path), mapped_path_quiet(root))
end

defp shared_download_root?(_save_path, _client_info), do: false
```

That value is optional config:

```elixir
|> put_if_present(:download_directory, System.get_env("#{prefix}DOWNLOAD_DIRECTORY"))  # config/loader.ex:318
```

If an operator never set `DOWNLOAD_CLIENT_<N>_DOWNLOAD_DIRECTORY`, the catch-all clause returns `false`, the `cond` falls to its third branch, and `do_save_path_fallback/2` sweeps the directory exactly as before. The only log line emitted is the existing info-level `"Found files via save_path fallback"`, which says nothing about the guard having been skipped for lack of configuration. So on an unconfigured instance the fix silently degrades to the old behaviour and there is no way to tell from the logs.

This compounds with the qBittorrent path. `content_files/2` deliberately returns `nil` when `content_path == save_path` (multi-file torrent added with subfolder creation disabled), which is a sound decision on its own, but it hands off to the fallback. **Both new code paths terminate at the same guard**, so a missing `download_directory` re-opens the original hole for precisely the shape that caused the incident.

Two options, not mutually exclusive:

1. **Source the root from the client instead of trusting operator config.** Transmission exposes it via `session-get` (`download-dir`), qBittorrent via `/api/v2/app/preferences` (`save_path`). Both are one request and would make the guard self-configuring.
2. **At minimum, log a warning when the fallback runs and `download_directory` is nil or empty**, so the degraded case is visible. Something like "save_path fallback running unguarded: no download_directory configured for client X".

For reference, on the instance that hit this, `DOWNLOAD_CLIENT_1_DOWNLOAD_DIRECTORY=/mnt/storage/media/Downloads` is set and matches Transmission's `download-dir` exactly, so the guard does fire there. That download root currently holds **534 entries**, versus the 38 in the rqbit root that produced 913 stray files and ~980 GiB of duplication. The downside if the guard ever fails to engage is now considerably larger than what we just cleaned up, which is why I would not leave it dependent on an optional env var.

## Not blocking, for follow-ups

**The `download.episode` fallback is untouched.** `media_import.ex` still logs `"Episode not found in database, falling back to download episode"` and then binds the file to `download.episode` with a destination from `build_destination_path/2`. With scoping fixed this is far less dangerous, since the file set is now genuinely this download's, but it remains a path where an unparseable file is attached to whatever episode the download row names.

**Separately, and I think higher value than anything left here: `media_files` has no uniqueness constraint.** `20251104023003_create_media_files.exs` created `unique_index(:media_files, [:path])`. When `20251115185757_make_media_files_path_nullable.exs` moved identity to `relative_path` + `library_path_id`, its `up` dropped that index and recreated only three non-unique indexes; the `down` restores it, so this looks like collateral rather than intent. Nothing has enforced uniqueness since.

Every duplication symptom traces to that gap: 1233 rows over 694 files during the incident, 615 rows over 330 paths when a library scan and a manual import both ran, and 153 duplicate paths still sitting there from two scans back in February and March. A `unique_index(:media_files, [:library_path_id, :relative_path])` plus a dedupe migration would have capped the blast radius of this very bug, and turns silent duplication into a loud constraint error. Worth its own PR.

PR #239, opened 2026-07-29
fix(player): one cast control surface, reachable from any screen

arsfeld:
## Problem

Casting rendered two UIs at once. `PlayerScreen` swapped its whole body for a full-screen remote whenever a cast was active, while `CastMiniController` — mounted in the global `MaterialApp.builder` stack — overlaid *every* screen, including the one already showing the remote. Both surfaces displayed the same four things (title, "Casting to <device>", play/pause, stop), and the bar sat on top and clipped the remote's "Stop Casting" button.

Separately, the only cast button in the app lived inside `PlayerScreen`. You had to start local playback before you could hand it off — there was no way to pick a device up front and have playback simply begin on the TV.

## What this does

**One control surface.** The full-screen remote is deleted. `CastMiniController` becomes the sole place casting is controlled, gaining a draggable scrubber, ±10s skips, play/pause, stop-with-confirmation, an idle state, and the stale-session Reconnect/Stop that used to live in the remote. `PlayerScreen` now shows an inert placeholder while casting: icon, "Playing on <device>", title, back button.

**Casting from anywhere.** A new `castTargetProvider` records a device chosen before anything is playing — a preference, not a connection, so `CastSessionManager` keeps owning exactly one concept (an active session with media). `PlayerScreen` reads it before initialising and calls `startCast` instead, never constructing a local player. All seven playback entry points route through `/player/:type/:id`, so this is one interception rather than seven call-site edits.

The affordance itself is an overlay pinned inside `AppShell`'s `Stack`, not an app-bar action: on desktop the browse screens deliberately suppress their app bars (`HomeScreen` passes `appBar: null`; Unwatched/Favorites/RecentlyAdded/Collections return a zero-height `SizedBox.shrink()`), so an app-bar-only button would have been invisible on macOS. Screens that *do* keep an app bar with actions — Library, Downloads, Settings, Search, and the three detail screens — get the button there instead, with `AppShell.hasOwnCastButton` gating the overlay so every route has exactly one affordance.

**Receiver duration fix** (first commit, independent of the UI work). A Chromecast reports `duration: -1` for any stream whose length it cannot determine — which is every HLS playlist Mydia serves, because the server runs FFmpeg without `-hls_playlist_type` on purpose so playback can start before the file is fully remuxed, leaving no `#EXT-X-ENDLIST`. `dart_cast` forwarded that `-1` verbatim. It rendered the scrub bar as `00:-1`, made the slider seek to `fraction × -1s` (the start of the video), and made skip-ahead clamp against a `-1s` "end" — also the start. The placeholder is now stopped at the backend boundary, the session is seeded with the runtime the app already knows, and the shared arithmetic lives in `cast_seek.dart` so every consumer treats non-positive as unknown.

## Testing

732 tests pass; 0 analyzer errors or warnings. New coverage includes the bar's three states, `--:--` with a disabled scrubber on unknown duration, seek-on-release (asserting zero seeks *during* a drag and exactly one on release), the overlay's route gating, and `CastButton`'s idle/casting states.

## Known gaps

- `_castToTargetIfSet` has no test — `player_screen.dart` has no scaffold, and one needs GraphQL, auth, downloads and cast providers plus a `media_kit` Player. Extracting it as a free function over `(Ref, fileId, …)` is the cheap way in if we want the clear-on-success contract pinned.
- `CastButton`'s black-circle styling now appears in four differently-styled app bars, including the light Settings and Search bars where it will look out of place.
- Not yet verified on a real receiver; the flow needs a manual pass on hardware.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,4 @@ docs/superpowers/
 
 # Subagent-driven-development scratch (ledger, briefs, review packages)
 .superpowers/
+player/ios/scripts/test-ci-signing.sh
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -28,20 +28,20 @@
       }
     },
     "../deps/phoenix": {
-      "version": "1.8.1",
+      "version": "1.8.8",
       "license": "MIT",
       "devDependencies": {
-        "@babel/cli": "7.28.3",
-        "@babel/core": "7.28.3",
-        "@babel/preset-env": "7.28.3",
-        "@eslint/js": "^9.28.0",
+        "@babel/cli": "7.28.6",
+        "@babel/core": "7.29.0",
+        "@babel/preset-env": "7.29.3",
+        "@eslint/js": "^10.0.1",
         "@stylistic/eslint-plugin": "^5.0.0",
         "documentation": "^14.0.3",
-        "eslint": "9.34.0",
-        "eslint-plugin-jest": "29.0.1",
+        "eslint": "10.2.1",
+        "eslint-plugin-jest": "29.15.2",
         "jest": "^30.0.0",
         "jest-environment-jsdom": "^30.0.0",
-        "jsdom": "^26.1.0",
+        "jsdom": "^29.0.1",
         "mock-socket": "^9.3.1"
       }
     },
@@ -49,10 +49,10 @@
       "version": "4.3.0"
     },
     "../deps/phoenix_live_view": {
-      "version": "1.1.16",
+      "version": "1.2.1",
       "license": "MIT",
       "dependencies": {
-        "morphdom": "2.7.7"
+        "morphdom": "2.7.8"
       },
       "devDependencies": {
         "@babel/cli": "7.27.2",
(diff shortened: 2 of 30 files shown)


arsfeld:
Addressed the Copilot review in c482dd1a — tooltips on all five `IconButton`s in `CastMiniController` (idle clear, rewind, play/pause, forward, stop).

Worth noting why this mattered more than the usual accessibility nit: this PR *deletes* the full-screen cast remote, so the bar is now the only cast surface in the app. An unlabelled icon here left a screen-reader user with no route to the control at all, rather than a worse route. The stale-session Reconnect/Stop buttons already carry visible text labels, so they needed nothing.

Unrelated: the earlier `Build / Packages` failure was infrastructure, not this branch — GitHub's Actions cache returned `ResourceExhausted: rate limit exceeded`, so the Nix binary cache failed every download with HTTP 418. This push re-triggers it.

arsfeld:
Correction on `Build / Packages`: the third run got past the cache throttling and surfaced a **real** failure that the earlier rate-limit noise was masking.

```
error: hash mismatch in fixed-output derivation '…-npm-deps.drv':
  specified: sha256-uphoD/oJoGADVR6cTdW+Pah4pLkCS1cKFc5Dp5ZAY4o=
     got:    sha256-hkcy/pcrWC8dtDwx+9wmHPAiI3rT9eZTQfIxafkztos=
```

This branch's package-lock refresh (5a6d4fa4, phoenix 1.8.8 / live_view 1.2.1) changed the npm dependency set, so the `npmDepsHash` pinned in `nix/packages/flake-module.nix` no longer matched. Fixed in 59bed3fb.

My earlier comment called this failure infrastructure-only. That was wrong — accurate for the first two runs, which never reached this derivation, but the wrong conclusion.

PR #266, opened 2026-07-30
fix(quality): apply the default quality profile to items with no profile

arsfeld:
> **Reworked onto `master`.** While this branch was in flight, #258 landed `Mydia.Indexers.QualityProfileResolver` with the same core insight and the same three call sites. Rather than ship a second resolver, this branch now **adopts master's** and keeps only what master does not have. The duplicate `Settings.effective_quality_profile/1` and its wiring have been deleted.

## What master already fixed (#258)

Profile-less items now resolve through `QualityProfileResolver` (item's profile → instance default → nil, logged), wired into the movie job, the TV job, and the manual search dialog. Master's `modals.ex` gate reads the profile back out of `@manual_ranking_opts`, so the score display and the ranking cannot disagree. That approach is kept as-is here.

### 1. The ranker preferred the *lowest* resolution (the important one)

`ReleaseRanker.sort_by_score_and_preferences/2` sorts by `{quality_index, -score}`, so a resolution's **position** in `:preferred_qualities` is the primary key and score only breaks ties. Every profile stores that list **ascending**: the shipped "Any" preset is `["360p", "480p", ..., "2160p"]`, and the Admin editor renders its checkboxes ascending while HTML array params submit in DOM order.

So any multi-resolution profile made the ranker grab the *lowest* resolution available. The repo's own test says so out loud: *"720p should come first even if 1080p has higher base score."*

**#258 made this more reachable without fixing it.** Profile-less items now fall back to the instance default, so any operator with a configured multi-resolution default is currently getting lowest-resolution-first grabs on master.

Fixed by sorting `:preferred_qualities` descending in `RankingOptions.build_quality_options/2`, sourcing the canonical order from the new `QualityProfile.valid_resolutions/0` rather than a duplicated literal. Unknown resolutions sort last without raising, and `RankingOptions.build/1` stays free of database access.

Note this is a deliberate behaviour flip: an operator who checked 720p and 1080p previously got 720p and will now get 1080p.

`release_ranker_test.exs:258` is untouched and still passes. It calls `rank_all/2` with a hand-ordered list, so it pins the ranker's own contract; this fix sits upstream of it.

### 2. Fresh installs seed a default

`ensure_default_quality_profiles/0` now sets `media.default_quality_profile_id` to "Any", but **only when the config row is absent entirely**. `set_default_quality_profile(nil)` writes `value: ""` rather than deleting the row, so an operator who deliberately cleared the default is never overridden on boot. A failed seed now logs instead of silently no-opping.

### 3. The blank dropdown options say what they do

`nil` means "follow the default", so the blank options now read "Use default (‹name›)" with the actual profile name, and the Admin selector reads "None (no default)". "Any Quality" was doubly wrong: it never selected the "Any" profile, it selected no profile.

The inert batch-edit clear option was removed rather than relabelled, because `index.ex:1018` and `media.ex:496` both discard it, so it never cleared anything. Making it work is a separate feature.

### 4. Import no longer freezes the default onto every item

`metadata_enricher` stamped the current default onto every auto-created item, so items kept a stale profile when the operator later changed the default. Removed: `nil` now carries the right meaning and resolves at search time. The Add Media toolbar pre-selection is untouched, since that is an explicit user choice.

### 5. `LIBRARY_PATH_<N>_QUALITY_PROFILE_ID` removed

It never worked: declared `:integer` while profile IDs are `binary_id` UUIDs, and read by nothing. The column and association stay, so no migration.

This removal also fixes a worse latent bug. The variable's own documented example value was `1`, and an integer *did* parse and pass validation, then failed the `binary_id` cast in `LibraryPath.changeset/2`. `Repo.update` returned `{:error, changeset}`, which `sync_from_runtime_config/0` only logs as a warning, so an operator following the docs had that library path silently fail to sync **entirely**.

### 6. Docs

`quality-profiles.md` documented the per-library mechanism that never worked and is now deleted, and never documented the real one. Rewritten around the two real tiers.

## Verification

`mix precommit` green on all 7 steps, including `credo --strict`. 5705 tests, 0 failures, 34 skipped.

**The manual UI smoke check has not been run** (no browser available). Worth doing before merge: on a seeded database, confirm an unstamped item's manual search puts a 1080p release above a 720p one, and that the profile dropdown reads "Use default (Any)".

## Known follow-ups

- Batch-editing has no clear-to-default option (the old one never worked, so it was removed rather than left as a false promise).
- `library_path_sync.ex:172,185` still pass a now-provably-nil `quality_profile_id` into two changesets. Harmless dead pass-through.
- The resolution vocabulary is still duplicated as a literal in four places in `admin_quality_profiles_live/components.ex`. The ranking path is now single-sourced, but the admin UI can still drift from `QualityProfile.valid_resolutions/0`.


arsfeld:
Copilot generated no inline comments, but its one suppressed low-confidence note was a real bug and is fixed in 9b649f4f.

`set_default_quality_profile(nil)` returned `{:ok, nil}` without writing a row when none existed. The seeding logic keys off that row's *absence* to mean "never configured", so an operator who explicitly chose "None (no default)" while no row existed would be silently re-seeded with "Any" on the next boot.

It is narrow but reachable: delete the "Any" profile, and boot seeding writes no row at all (`Repo.get_by(QualityProfile, name: "Any")` returns nil). Pick "None" at that point and nothing is persisted. `ensure_default_quality_profiles/0` then recreates "Any" on the next boot and seeds it, overriding the choice.

Clearing now persists a row with an empty value. Readers are unaffected, since `get_default_quality_profile_id/0` already returns nil for an empty value. Added a regression test covering that exact path.

PR #299, opened 2026-08-02
ci(release): pin releases to a commit and refuse to publish partial builds

arsfeld:
Audit of the release process, plus fixes for what it turned up.

## What was wrong

**A release could publish while broken.** `publish` ran under `if: always()` and only aborted when *zero* assets existed. A failed Windows build published a release silently missing its installer. Each platform job also had a `check_<platform>` step that skipped the entire build when the directory was absent and reported success either way. All six directories exist under `player/`, so those gates were vestigial from when platforms were being added one at a time, and their only remaining effect was letting a job report success while building nothing.

**The git tag could point at code nobody built.** The draft targeted `master`, a floating branch name. That got resolved once when the jobs checked out `github.sha`, and again when GitHub created the tag at publish. Docker builds take up to two hours, so any merge in that window left the tag on a different commit than the images. Confirmed on the live `v0.13.0-beta.1` draft, whose `target_commitish` was the literal string `"master"`.

**`:latest` could move backwards.** `docker-merge` applied `latest`, `<major>` and `<major>.<minor>` to any stable release, so a v0.11.2 patch after v0.12.0 would repoint all three at the older build. Same for `:beta` across prereleases.

**Nothing rehearsed the release path.** Six of the last fourteen completed release runs failed, and `release.yml` has taken 25 commits in six months, mostly `fix(ci)` / `fix(release)`. Signing, notarization, appcast generation, Inno Setup, TestFlight and Play Store upload only ever executed during a real release, so every break was found while shipping.

**The documented process did not exist.** `.claude/commands/release.md` still described the pre-June tag-first flow. Following it creates a premature tag and a non-draft release that `prepare` rejects with exit 1. There was no release documentation anywhere in `docs/`.

## What changed

- **A release names a commit.** The draft must target a full SHA. Every job checks it out; Docker records it as `BUILD_COMMIT` and the `revision` label. `prepare` also lists commits that landed after the draft was cut and stops, since the notes were written without them. `accept_drift` ships anyway.
- **Publish blocks on any failed platform.** `allow_missing` waives one by name for a genuine outage, and the waiver is appended to the release body so the gap is visible on the release page. The `check_<platform>` gates are gone.
- **Floating tags only move forward**, compared with `sort -V` against published releases of the same class.
- **`dry_run` rehearses everything** that has historically broken, without pushing to ghcr, uploading to the stores, publishing, or deploying docs.
- **`Player / Web` removed.** It duplicated the Dockerfile's `flutter-builder` stage, which already bakes the web player into `priv/static/player/`. Its artifact was downloaded by nothing while gating `publish`.
- **Docs**: `/release` rewritten, and `docs/development/releasing.md` added, covering both this flow and metadata-relay's separate tag-push flow.

## New dispatch inputs

| Input | Default | Effect |
| --- | --- | --- |
| `version` | required | Tag of an existing draft release. Ignored when `dry_run` is set. |
| `dry_run` | `false` | Build, sign and notarize without publishing, pushing, or uploading to stores |
| `accept_drift` | `false` | Proceed although master moved past the draft's commit |
| `allow_missing` | `""` | Comma-separated platforms whose failure must not block publish |

## Verification

Workflow changes cannot be fully proven without a run, so this is what was actually checked:

- `actionlint` reports no expression or syntax errors. Findings dropped 43 → 28, all remaining ones pre-existing shellcheck style notes in untouched scripts.
- The publish gate's shell was **extracted from the committed workflow** and run against 7 scenarios: all-green, single failure, waived failure, two failures with one waived, `skipped` treated as failure, waivers with mixed case and whitespace, and a waiver naming a job that passed. All behave correctly. Two bugs were found and fixed this way: `IFS=', '` only joins on the comma, and the suggested `allow_missing` dropped platforms already waived, so pasting it would have silently un-waived them.
- A step-name diff against `master` confirms the only removals are `Check platform support` from each platform job and the `player-web` job. Every build, sign, notarize and upload step survives.
- `mkdocs build --strict` passes with the new page in the nav.

**Not verified:** no release or rehearsal has been dispatched with this workflow yet. The intended next step is `gh workflow run release.yml -f dry_run=true` on this branch before the next real release uses it.

## Follow-ups, not in this PR

- `v0.13.0-beta.1` published successfully under the old workflow while this PR was being written, so it needs no action. Its `target_commitish` is still the literal `"master"`, which is the bug this PR closes for future releases.
- The rehearsal stops before the store uploads by choice, so App Store Connect and Play Store credential expiry still surfaces first during a real release.
- `docs/development/releasing.md` matches master's layout. The unmerged `docs/diataxis-restructure` branch moves these pages to `docs/contributing/`, so one of the two will need to relocate it on merge.

### DIFF
--- a/.claude/commands/release.md
+++ b/.claude/commands/release.md
@@ -1,62 +1,99 @@
 ---
 allowed-tools: Bash(git*,gh*), Read, Edit
 argument-hint: [major|minor|patch|beta|rc (optional - auto-detected)]
-description: Create and publish a new release with version bump and release notes
+description: Cut a draft release pinned to a commit and dispatch the release workflow
 ---
 
 # Release Process
 
-Create a new release by:
+Releases are draft-first. You create a draft pinned to a commit, then dispatch
+`release.yml`, which builds every platform, uploads the assets, and publishes
+the draft. **Never create a git tag yourself.** GitHub creates it when the
+workflow publishes the draft.
 
-1. **Analyze changes since last release**
+Full reference: `docs/development/releasing.md`.
 
-   - Get the latest tag using `git describe --tags --abbrev=0` or `git tag --sort=-version:refname | head -n1`
-   - Use `git log <last-tag>..HEAD --oneline` to see all commits since last release
-   - Look at the git diff and commit messages to understand the nature of changes
+## 1. Analyze changes since the last release
 
-2. **Determine version increment** (unless specified as $1)
+- Latest tag: `git tag --sort=-version:refname | grep -v metadata-relay | head -n1`
+- Commits since: `git log <last-tag>..origin/master --oneline`
+- Read the commit messages and diffs to understand what actually changed
 
-   - **PREFER patch (0.0.X)** for bugfixes, small improvements, or unclear changes
-   - Use **minor (0.X.0)** ONLY for significant new features or major functionality additions
-   - Use **major (X.0.0)** ONLY if explicitly instructed with `$ARGUMENTS` containing "major"
-   - Use **beta** to create a pre-release version with `-beta.N` suffix (e.g., v1.2.3-beta.1)
-   - Use **rc** to create a release candidate with `-rc.N` suffix (e.g., v1.2.3-rc.1)
-   - Current version is derived from the latest git tag (mix.exs reads BUILD_VERSION at compile time)
-   - Parse the latest tag version, increment appropriately
+## 2. Determine the version increment (unless given as `$1`)
 
-3. **Categorize changes** for release notes:
+- **Prefer patch (0.0.X)** for bugfixes, small improvements, or unclear changes
+- **minor (0.X.0)** only for significant new features
+- **major (X.0.0)** only when explicitly asked
+- **beta** for a prerelease with a `-beta.N` suffix
+- **rc** for a release candidate with an `-rc.N` suffix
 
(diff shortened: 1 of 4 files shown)


arsfeld:
Both findings addressed in f51a4d42.

**Dead doc reference** — real bug. The header pointed at `docs/contributing/releasing.md`, but the page is at `docs/development/releasing.md` on master. (`docs/contributing/` exists only on the unmerged `docs/diataxis-restructure` branch, which is where I got the path from.) Fixed.

**Hardcoded `master`** — taking this one, though not for the stated reason. This workflow isn't reusable and the default branch here is `master`, so the fork scenario doesn't apply. The real problem is internal: the drift check compares against `origin/${DEFAULT_BRANCH}` while every message describing that comparison said "master", and two of them printed a fix command against `origin/master`. That's my own inconsistency, and interpolating costs nothing. All five runtime messages now name the branch actually compared.

The example in the header comment keeps the literal `origin/master` deliberately: it's meant to be copy-pasted into a shell, where `${DEFAULT_BRANCH}` would expand to nothing. Same reasoning for the concrete commands in `docs/development/releasing.md` and `.claude/commands/release.md`.

Verified the escaping still renders a copy-pasteable command: `--target $(git rev-parse origin/main)` with the branch interpolated and the `$(...)` left literal.

PR #302, opened 2026-08-02
fix: wire up the backups, grants, and health checks the docs audit found missing

arsfeld:
Fixes the code bugs surfaced by the documentation audit in #300. Stacked on that branch, so review it after #300 or merge #300 first.

Each of these is a feature that exists in the codebase but does not reach the operator. That is why an audit of the docs found them: the docs described the intended behaviour, and the code did not implement the last step.

## 1. Pre-migration backups never ran (`6f207540`, `7a31e06a`)

`Release.backup_before_migrations/0` creates a timestamped backup when migrations are pending. It works, it is tested, and its only caller was a mix task invoked from `devenv.nix`. Production runs migrations from the supervision tree with no backup step, and `skip_migrations?/0` returns true unless `RELEASE_NAME` is set, so migrations ran precisely in the container case that had no backup.

Self-hosters upgrade by pulling an image and restarting. If a migration went wrong, there was nothing to restore.

**Now:** a `Mydia.Release.MigrationBackup` child sits between `Mydia.Repo` and `Ecto.Migrator`, takes the snapshot, and returns `:ignore`. `nix/module.nix` runs migrations via `ExecStartPre` and bypasses the supervision tree entirely, so `migrate/0` covers that path too.

**Snapshots use `VACUUM INTO`, not a file copy.** Production SQLite runs in WAL mode, so `File.cp` can silently omit committed transactions still in `mydia.db-wal`. The first attempt added `PRAGMA wal_checkpoint(TRUNCATE)`, but that pragma reports `busy` when it cannot complete and the result was discarded, so under exactly the load this feature exists for it could still produce a short backup. `VACUUM INTO` reads through the WAL, is atomic, and cannot produce a torn file. A test writes rows that live only in the WAL and asserts the snapshot contains them.

**PostgreSQL is not silently "protected".** `get_database_path/0` reads `Repo[:database]`, which on Postgres is the database *name*, so the old code would have matched `is_binary` and failed with `{:database_not_found, "mydia"}`. It now branches on `Mydia.DB.sqlite?/0` and logs an actionable warning with the `pg_dump` command. No shellout: `pg_dump` is not guaranteed present, and an unreliable backup is worse than a clear "there is none".

**A failed backup logs at `:error` and startup continues.** Refusing to boot would lock an operator out over a full disk they cannot fix from inside the instance, and in a container that becomes a restart loop.

**`SKIP_BACKUPS` is now real.** It was documented but read nowhere. A boot-time snapshot of a multi-gigabyte database is a genuine cost.

## 2. Manifest revisions silently half-broke plugins (`3066b301`)

Revising a plugin manifest to request new capabilities updated the manifest but left the grant untouched. The plugin stayed approved and enabled while calls into the new capabilities returned `Denied`. The security posture was right; the feedback was missing. Operators saw a healthy plugin that quietly did not work.

**Now:** `Plugins.Capabilities.ungranted/2` compares requested against granted, per value for the capabilities the host gates by value (`events:subscribe`, `net:http`, `data:read`, `surfaces:write`) and per class for flags, failing closed on opaque payloads. The admin UI marks affected plugins and offers re-approval. One warning per activation names the ungranted capabilities; no per-call logging, since a denial can fire in a hot loop. Grants are still never widened without an explicit operator action.

**This also closed an adjacent hole.** `update_settings/2` recomputed the `net:http` grant from the *current manifest*, so saving an unrelated setting after a revision silently granted the newly declared static hosts and cleared the warning. It now derives the static side from the grant, so "never widened without approval" is actually true.

## 3. The stuck-upgrade health check was never surfaced (`ccc59106`)

`Health.upgrade_health/0` counts media files wedged mid-upgrade past a threshold. It had no caller outside its own test, so automatic quality upgrades shipped with the check written but wired to nothing.

**Now:** surfaced on `/admin/config/status`, hidden entirely at zero, and when non-zero it names the cause and links to the jobs page. Computed at mount rather than on the 5s refresh timer. The upgrades how-to gained the troubleshooting section it lacked.

## 4. Preset id uniqueness guard (`3b47d3d5`)

A test asserting every quality profile preset id is unique.

**This started as a reported duplicate-id bug and the report was wrong.** The audit grep counted `id:` occurrences inside `@doc` examples as preset data, producing "28 defined, 23 distinct". At runtime there are 23 presets with 23 distinct ids and always have been. Nothing was shadowed and nothing was renamed. The test stays because the class of bug is real and cheap to exclude.

## 5. Documentation and SDK comment corrections (`0916601c`)

- The SDK's doc comments named the attribute macro `#[mydia::plugin]`; the re-export is `mydia_plugin_sdk::plugin` and every real call site uses that. The WIT-generated `mydia::plugin::host` namespace is a different thing and is untouched.
- `AGENTS.md` described the pre-iroh design (libp2p-style mesh, Kademlia DHT, mDNS, TCP, Noise). None of it is present; the core is iroh 1.0 over QUIC with TLS 1.3 and DNS-based discovery. That stale section is where the same false claim in the published docs came from, which is why it is worth fixing rather than leaving.
- Documents that `SKIP_BACKUPS` trades away the WAL-aware snapshot, so a volume snapshot only substitutes for it when atomic across `mydia.db`, `-wal`, and `-shm`.

## Testing

289 tests, 0 failures across the touched suites (release, plugins, presets, admin status LiveView). `compile --warnings-as-errors`, `credo --strict`, and `mkdocs build --strict` all green.


arsfeld:
@copilot fix the conflicts

PR #323, opened 2026-08-04
feat(changelog): show release notes after an upgrade

arsfeld:
Shows an operator (and every user) what changed after the server is upgraded, without leaving the app and without network access.

## What this does

Release notes ship as one markdown file per stable release under `priv/changelog/`, compiled into the image at build time. A dismissible banner appears in the shared layout when a user has unread releases, and `/changelog` lists the full history with unread entries expanded.

A user who was away for several releases sees all of them stacked, newest first. A brand-new user, and every user that exists when this ships, silently adopts the current version and sees nothing, so nobody is greeted by 33 releases of history.

## Design decisions worth knowing

**Bundled, not fetched.** The notes for the version you are running are knowable at build time, which is exactly the on-upgrade case. So there is no runtime file IO, no network call, and no GitHub rate limit. It works on an air-gapped instance.

**No migration.** The per-user seen state is one key in the existing `user_preferences.preferences` map column, which is what that schema exists for.

**No build-version check anywhere.** Behaviour is a pure function of the bundled notes and the user's stored value. Development and prerelease builds fall out correctly as a consequence rather than through a special case: a `:master` image that bundles nothing new shows nothing.

**Each file holds only its own version's changes.** The app stacks releases itself, so duplicating a previous version's content would render it twice in adjacent cards.

## This changes the release process

Cutting a release gains one step, and it must be the last merge before you pin a commit, because the draft targets a SHA and the image is built from it:

1. Open a PR adding `priv/changelog/0.13.0.md`, then merge it.
2. Pin and cut the draft with `--notes-file priv/changelog/0.13.0.md`.
3. Dispatch as before.

The `prepare` job now fails when `priv/changelog/<version>.md` is missing at the pinned SHA, before any platform build starts. Prereleases are exempt and keep hand-written notes.

For a patch release, the GitHub notes are the patch file concatenated with the preceding minor's file. That makes the existing "patch releases carry the minor's notes" rule mechanical instead of something to remember, while each bundled file still holds only its own changes.

Both `docs/contributing/releasing.md` and `.claude/commands/release.md` (the `/release` runbook) are updated. The runbook mattered: it previously told you to write notes to a scratch `notes.md`, which would have failed the new guard on the very next release.

## Backfill

All 33 stable releases, v0.2.0 through v0.12.0, imported verbatim from their GitHub bodies. Two were corrected because they violated the one-file-one-version rule: `0.7.1.md` was a near-verbatim copy of `0.7.0.md`, and `0.11.1.md` embedded all of `0.11.0.md`. Thirteen files also had a `# Mydia vX.Y.Z` heading duplicating the version the card already renders.

## Verification

- Full precommit green: compile, `deps.unlock --unused`, `format --check-formatted`, `credo --strict`, and the suite at **6213 tests, 0 failures**.
- `mix assets.build` compiles the new scoped CSS.
- The compile-time guard was verified by hand: a malformed filename fails the build.
- The `@external_resource` directory tracking was verified empirically, 33 to 34 entries on adding a file and back on removing it, without `--force`.

Not verified: the release workflow was deliberately not dispatched, since a dry run builds and signs every platform. The guard was checked with `bash -n`, a YAML parse, and a path-by-path read of all four branches (stable with notes, stable without, prerelease, dry run).

The concurrency fix in `Mydia.Accounts` was exercised on SQLite locally. CI covers PostgreSQL.

## One pre-existing bug fixed along the way

`get_user_preference!/1` did get-then-insert with no unique constraint on the changeset, so concurrent callers raised `Ecto.ConstraintError`. That was mostly latent before, but this feature makes the function run on every authenticated LiveView mount, with the insert firing once per user on their first page load after an upgrade, which is precisely when a session restore reopens several tabs at once. It now tolerates the race and always re-reads the row.

## Known follow-ups, not blocking

- The regression test for that race attaches a VM-global telemetry handler. On the PostgreSQL CI job, where `DataCase` honours `async: true`, an unrelated test's `user_preferences` query can consume a barrier slot and silently disarm the guard. The degradation is false-green rather than flaky-red, so it will not cause spurious failures, and the production fix is correct regardless. Scoping the handler to the racer pids would harden it.
- `priv/changelog/0.7.1.md` keeps its original compare range `v0.6.0...v0.7.1`, which spans two releases. That is what the real GitHub release linked; left as-is rather than invented.
- `.claude/commands/release.md` does not yet say "no `# Mydia vX.Y.Z` heading", so a future release could reintroduce the duplicated version line.


arsfeld:
Thanks, addressed all three in fb0561ac. One correction on the first.

**Markdown escaping.** The premise is not quite right: `Earmark.as_html/1` escapes inline HTML by default. `escape: true` is set in `%Earmark.Options{}` (`deps/earmark/lib/earmark/options.ex:55`) and honoured in `transform.ex`, and the call passed no options, so it took that default. A `<script>` committed into a notes file already came out as `&lt;script&gt;`.

Worth stating the fuller picture, because it is the reason this was reviewed carefully rather than waved through: the HTML is a compile-time literal. `Path.wildcard/1` and `Earmark` both run during compilation, the result is baked into the BEAM, and there is no runtime file read. Nothing an HTTP request carries can reach that sink. And write access to `priv/changelog/` implies write access to `lib/**/*.ex`, which is arbitrary code execution in the built image, so injecting a script tag would be a strict downgrade in capability.

Made it explicit anyway. The output goes through `raw/1`, and pinning the flag means a future change to Earmark's default cannot silently start emitting raw HTML.

**Telemetry handler isolation.** Correct, and worth acting on: an independent review of this branch flagged the same thing, which is what moved it from "noted" to "fixed". Scoped the handler to the two racer processes via an opt-in flag rather than making the module synchronous, so the test stays isolated and still runs in the async phase. Also stopped it sending its own wakeup to itself, which left a stray message in the mailbox.

**Verified routes.** Agreed, and this one was a stale decision on my part. Those links were plain strings because `/changelog` did not exist yet when the banner was written, so `~p` would have been a compile error at that point. The route landed in a later commit and I never revisited it. Both now use `~p"/changelog"`. The file's other nav links stay plain strings, which is pre-existing and out of scope here.

PR #325, opened 2026-08-04
Remove the unauthenticated read_media p2p request

arsfeld:
Closes an unauthenticated arbitrary file read on the p2p transport.

## The problem

`read_media` accepted a peer-supplied `file_path`, checked `File.exists?`, and streamed the bytes back. `ReadMediaRequest` carried no `auth_token` field at all, unlike `GraphQLRequest` and `HlsRequest` which both do, so nothing in the path checked a credential. The accept loop validates ALPN and nothing else, and iroh publishes the node ID over DNS discovery, so reachability was not obscure. Any peer able to dial the node could read any file the server process could read.

The existing code comment said as much:

```elixir
# SECURITY: In production, verify path is within allowed directories!
```

**Blast radius is bounded to installs that opted into remote access.** `remote_access_enabled` defaults to `false` (`config/config.exs`), and `Mydia.P2p.Server` only starts when that flag is on. Operators running with remote access enabled should upgrade.

## Why delete rather than secure it

Nothing calls it. Verified across `.rs`, `.dart`, `.ex` and `.exs`: the only references were the type definitions, the NIF dispatch arm, and the handler. Git history confirms it was never live in a shipped client, `a0799ba8` removed the last client-side caller and replaced it with `notImplemented`.

Offline download, the one feature that would plausibly want raw file transfer over p2p, already runs on a safer path: the player requests a job id over authenticated GraphQL, then fetches via an HLS request carrying an auth token with `session_id: "download:{jobId}"`, and the server resolves the path from the job rather than accepting one from the client. That is the capability-lookup pattern, strictly better than confining a client-supplied path. Securing `read_media` would have meant maintaining a second, weaker way to do what already works.

## Changes

Two commits, split so the tree stays both compilable and loadable in between:

1. Remove the request path: CBOR enum variant, `ReadMediaRequest`, `ElixirReadMediaRequest`, the NIF dispatch arm, `Mydia.P2p.ReadMediaRequest`, and the GenServer handler.
2. Remove the orphaned `respond_with_file_chunk` NIF and its Elixir stub, plus the two `std` imports it was the last user of. These two must land together: the BEAM refuses to load a native library registering a NIF with no matching Elixir function, which fails at load rather than at compile.

## Compatibility

No window to manage, which matters because this is self-hosted with no coordinated deploy order between server and player.

`serde_cbor` 0.11 with a plain derive uses externally tagged enums keyed by variant **name** (`{"ReadMedia": {...}}`), not by ordinal. Confirmed against the vendored crate source: `Serializer::new` defaults to `packed: false`, and `serialize_newtype_variant` writes a one-entry map keyed by the variant name. Removing this variant therefore leaves `Ping`, `Pairing`, `GraphQL`, `HlsStream` and `Custom` byte-identical on the wire. The fix is server-side only; no client change is needed and none helps.

## Verification

- `./dev mix precommit` passed (compile, unused deps, format, credo --strict, tests).
- Full suite: `6213 tests, 0 failures, 34 skipped (142 excluded)`.
- NIF registration confirmed at runtime with the app booted: module loads, `function_exported?(Mydia.P2p, :respond_with_file_chunk, 5)` is `false`.
- All 10 remaining `#[rustler::nif]` functions audited 1:1 against their Elixir stubs; no orphan in either direction.

## Deliberately out of scope

Three p2p authorization gaps remain, all pre-existing and none a regression from this PR:

- The accept loop still admits any dialer after the ALPN check (no pairing gate).
- `handle_download_stream/4` streams `job.output_path` with no check that the job belongs to the authenticated user. Requires a valid token, and installs are typically single-household.
- `MydiaRequest::Custom` is reachable without a credential. It reaches no handler and discloses nothing, but each request holds a pending-response entry, a task, and an open QUIC stream for the 30 second timeout.

All three are tracked for a follow-up focused on connection-layer authorization.

### DIFF
--- a/lib/mydia/p2p.ex
+++ b/lib/mydia/p2p.ex
@@ -46,13 +46,6 @@ defmodule Mydia.P2p do
   def send_response(_resource, _request_id, _response), do: :erlang.nif_error(:nif_not_loaded)
 
   @doc """
-  Read a file chunk and send it as a response.
-  This is optimized to read and send directly from Rust.
-  """
-  def respond_with_file_chunk(_resource, _request_id, _file_path, _offset, _length),
-    do: :erlang.nif_error(:nif_not_loaded)
-
-  @doc """
   Get network statistics including connected peers and relay status.
   """
   def get_network_stats(_resource), do: :erlang.nif_error(:nif_not_loaded)
--- a/lib/mydia/p2p.ex
+++ b/lib/mydia/p2p.ex
@@ -114,19 +114,6 @@ defmodule Mydia.P2p.PairingResponse do
         }
 end
 
-defmodule Mydia.P2p.ReadMediaRequest do
-  @moduledoc """
-  A request to read a media file chunk.
-  """
-  defstruct [:file_path, :offset, :length]
-
-  @type t :: %__MODULE__{
-          file_path: String.t(),
-          offset: non_neg_integer(),
-          length: non_neg_integer()
-        }
-end
-
 defmodule Mydia.P2p.NetworkStats do
   @moduledoc """
   Network statistics from the p2p host.
--- a/lib/mydia/p2p/server.ex
+++ b/lib/mydia/p2p/server.ex
@@ -285,26 +285,6 @@ defmodule Mydia.P2p.Server do
     {:noreply, state}
   end
 
-  def handle_info({:ok, "request_received", "read_media", request_id, req}, state) do
-    # Validate file path exists
-    # SECURITY: In production, verify path is within allowed directories!
-    if File.exists?(req.file_path) do
-      # Use the optimized NIF to read chunk and respond
-      P2p.respond_with_file_chunk(
-        state.resource,
-        request_id,
-        req.file_path,
-        req.offset,
-        req.length
-      )
-    else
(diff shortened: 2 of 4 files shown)


arsfeld:
Superseded by #695, which closes #687.

Your analysis here was right and is carried forward: nothing constructs a `ReadMedia` request, so refusing it beats building server-side path resolution for a capability with no client. I've since confirmed that against the full history (`git log --all -S'read_media'` scoped to `player/` returns zero commits) and found that mydia-rs already answers the request with an error of its own, calling it a legacy direct read path.

Closing rather than rebasing because this diff targets the handler as it existed before `serve_request/5` and the `RemoteAccess.enabled?()` gate landed, so it no longer applies.

Two things #695 does differently:

- The refusal lives in `mydia_p2p_core` rather than deleting the Elixir handler alone, so the request never crosses the NIF boundary and a loopback test can actually reach it. There's now a regression test that stands up two real hosts and asserts the refusal.
- `MydiaRequest::ReadMedia` stays on the wire. Deleting the variant, as this PR does, makes the request fail to decode, and the decode-failure path drops the stream without a response, so a peer would hang for 30s rather than being told no.

PR #330, opened 2026-08-04
fix(downloads): stop a titleless release crashing DownloadMonitor

arsfeld:
## Problem

`Mydia.Jobs.DownloadMonitor` crashed with:

```
** (FunctionClauseError) no function clause matching in String.downcase/2
    String.downcase(nil, :default)
    lib/mydia/downloads/torrent_matcher.ex:673: has_sequel_marker?/1
    lib/mydia/downloads/torrent_matcher.ex:296: calculate_movie_confidence/2
```

`ReleaseParser.infer_media_type/4` types a release as `:movie` on a year or quality token without consulting the title, and `ReleaseIntake` only validated the title for `:unknown` results. A titleless `:movie` therefore reached matching, where `has_sequel_marker?/1` was the one scoring function missing the `nil` clause its neighbours have (`normalize_string/1` has had one all along, which is why the other four call sites tolerate it).

Two real names reproduce it, both verified against the parser:

- `[Tracker] (2019) 2160p UHD BluRay x265-GROUP` - the validator runs on the original name where the bracket tag counts as meaningful content, then `clean_torrent_name/1` strips it before parsing, leaving no title. Tracker-tagged names are common, which is likely how this reached production.
- `1080p.BluRay.x264-GROUP` - the release group satisfies the validator, the parser then strips it as a release group.

## Blast radius

`find_and_match_untracked/0` had no per-torrent isolation, so one bad torrent aborted `DownloadMonitor.perform/1` at the untracked-matching step, killing the two passes that run after it: `ExternalTorrents.refresh/0` and stuck-download detection. The torrent stays in the client across runs, so the job failed on every invocation until an operator removed it by hand. Oban retries re-ran the earlier passes, which are idempotent, so no data was corrupted.

`ExternalTorrents.refresh/0` reaches the same scoring function through `find_top_candidates_in/3`. That path survived only because `Classifier.suggestions/2` already rescues, and that rescue logged a bare warning and reported nothing, so the same defect has been firing there unobserved.

## Changes

- Guard `has_sequel_marker?/1` against a nil title, matching the `normalize_string(nil)` convention already in the module
- Reject titleless releases at the `ReleaseIntake` boundary with a new `:missing_title` reason, kept distinct from `:unable_to_parse` so the two causes stay separable in logs
- Contain per-torrent exceptions in `UntrackedMatcher`, mirroring `MetadataRefresh.safe_refresh/2` including its load-bearing ErrorTracker report
- Report from the `Classifier.suggestions/2` rescue so it stops masking this class of bug

## Testing

New coverage for the nil-title guard (both with and without a year), both intake rejections plus two positive cases confirming normal movie and TV releases still parse, and the exception isolation (contained, and the happy path still works through the same entry point).

`mix precommit` green: 6225 tests, 0 failures.

## Note

An earlier draft of this change also filtered non-binary entries out of `alternative_titles`. It was dropped after testing showed it is unreachable: a nil alternative title scores 0.0 through `normalize_string/1` and `Enum.max_by` returns the first max, and the primary title is always a non-nil binary, so a nil variant can only win when the torrent title is also nil, which the first fix already covers.

### DIFF
--- a/.github/workflows/ci-player-e2e.yml
+++ b/.github/workflows/ci-player-e2e.yml
@@ -14,7 +14,10 @@ jobs:
   test-player-e2e:
     name: Test / Player E2E
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # A run that has to populate a cold layer cache still pays the full build
+    # cost, so this has to clear a cold build, not a warm one. 30 minutes did
+    # not, and the job died mid-`apt-get` without ever running a test.
+    timeout-minutes: 60
 
     steps:
       - name: Checkout
@@ -67,9 +70,27 @@ jobs:
         if: steps.should.outputs.run == 'true'
         uses: docker/setup-buildx-action@v3
 
+      # Built through bake rather than `docker compose build` so the layers can
+      # be cached across runs. The player test image installs the whole GTK /
+      # mpv / Flutter Linux desktop toolchain in one apt layer, which alone has
+      # taken ~20 of the job's 30 minutes and timed the job out before the tests
+      # ever ran. `type=gha` cache needs ACTIONS_CACHE_URL and
+      # ACTIONS_RUNTIME_TOKEN in the environment, which the Docker actions
+      # inject and a plain `run:` step does not, so this cannot be done as a
+      # shell step. Scopes are per-image so the three do not evict each other.
       - name: Build E2E images
         if: steps.should.outputs.run == 'true'
-        run: DOCKER_BUILDKIT=1 docker compose -f compose.player-e2e.yml build
+        uses: docker/bake-action@v6
+        with:
+          files: compose.player-e2e.yml
+          load: true
+          set: |
+            relay.cache-from=type=gha,scope=player-e2e-relay
+            relay.cache-to=type=gha,mode=max,scope=player-e2e-relay
+            mydia.cache-from=type=gha,scope=player-e2e-mydia
+            mydia.cache-to=type=gha,mode=max,scope=player-e2e-mydia
+            test.cache-from=type=gha,scope=player-e2e-test
+            test.cache-to=type=gha,mode=max,scope=player-e2e-test
 
       - name: Run player E2E tests
         if: steps.should.outputs.run == 'true'
--- a/.github/workflows/ci-player-e2e.yml
+++ b/.github/workflows/ci-player-e2e.yml
@@ -14,9 +14,10 @@ jobs:
   test-player-e2e:
     name: Test / Player E2E
     runs-on: ubuntu-latest
-    # A run that has to populate a cold layer cache still pays the full build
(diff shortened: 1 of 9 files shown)


arsfeld:
Addressed both suppressed review comments in 36b71b6a.

**`untracked_matcher.ex` — rescue could raise a second time.** Correct, and worth fixing even though it is not reachable through the current sole caller (torrents come from `Client.list_torrents` and then get `Map.put(:client_name, ...)` applied, so they are always maps by the time they arrive). The point stands regardless: this function's entire contract is that it never raises, so a rescue block that can itself raise defeats it. `Map.get/2` is now a total `torrent_field/2` helper, with a test passing `nil` straight in. The log confirms containment:

```
error=%BadMapError{term: nil} [error] Exception matching untracked torrent, continuing pass
```

**`release_intake.ex` — stale documented contract.** Also correct, and this one was a live documentation bug I introduced. The moduledoc explicitly enumerates the parse contract, and its final line still said `otherwise → {:ok, %ParsedFileInfo{}}`, which stopped being true the moment titleless releases started being rejected. Added the `:missing_title` case with the reason it is reachable.

Local suites green: 21 tests, 0 failures.

arsfeld:
Player E2E timed out twice at exactly 30 minutes, both times inside the `apt-get` layer of **Build E2E images**. The player tests never executed either time, so nothing in this PR was actually exercised and failed.

The cause is pre-existing and unrelated to this diff. `ci-player-e2e.yml` built three images with a bare `docker compose build` and no layer cache at all, so every run rebuilt the player test image from scratch. That image installs the whole GTK / mpv / Flutter Linux desktop toolchain in a single apt layer, which alone consumed roughly 20 of the 30 minute budget. The job passed on this branch's earlier commit `3e293d56` and the only delta since is 21 lines of Elixir and a moduledoc, which cannot affect Docker build time.

Fixed in 737b34f1 rather than re-run repeatedly:

- Builds through `docker/bake-action` instead of a shell step. This is not cosmetic: `type=gha` cache needs `ACTIONS_CACHE_URL` and `ACTIONS_RUNTIME_TOKEN`, which the Docker actions inject and a plain `run:` step does not, so the cache could not work from a shell invocation.
- Per-image cache scopes, so the three images do not evict one another.
- Explicit `image:` names on the compose services. The implicit tag derives from the compose project, which is the checkout directory, so CI and a git worktree disagreed on what the built image is called.
- `timeout-minutes` 30 to 60, because the run that populates a cold cache still pays the full build cost. Caching alone would not have made this run green.

Verified locally as far as is possible from outside CI: both files parse, and `docker compose config -q` validates. The caching behaviour itself can only be confirmed by this run.

Noting the hygiene tradeoff: a CI change does not really belong in a bugfix PR. It is here because it is what unblocks this one.

PR #369, opened 2026-08-08
fix(release): correct floating-tag selection and bound the iOS job

arsfeld:
Two independent faults the v0.13.0 release run exposed.

## 1. Floating Docker tags were silently skipped

The run pushed `:0.13.0` and `:0.13.0-pg` but moved no floating tag. `:latest`, `:0.13` and `:0` never updated, and `:latest` still resolves to the v0.12.0 digest.

`docker-merge` decides whether to apply floating tags by comparing this release against the newest already-published release of its class. That list is built from every release in the repository, stripping only a leading `v`:

    .tagName | ltrimstr("v")

This repository also publishes the metadata-relay under `metadata-relay-vX.Y.Z`. `ltrimstr("v")` does not match that prefix, so the tag survives intact, and `sort -V` places a string beginning with a letter above every numeric version. `NEWEST` resolved to `metadata-relay-v0.9.0`, and the job concluded:

    ##[warning]0.13.0 is older than the newest published stable release (metadata-relay-v0.9.0).
    Applying only :0.13.0 and skipping the floating tags so they keep pointing at metadata-relay-v0.9.0.

Filtering to Mydia's own `vX.Y.Z` tags fixes it. Applied to both the stable and prerelease branches: the prerelease path is correct today only because the metadata-relay releases are not marked prerelease, which is incidental rather than guaranteed.

**Why this did not surface earlier.** The `IS_NEWEST` guard was introduced in #299, during this release cycle. v0.12.0 predates it and applied floating tags unconditionally. The six v0.13.0 betas exercised only the prerelease branch. This run was the logic's first stable exercise.

**Severity.** Had the run reached publish, v0.13.0 would have shipped with `:latest` pointing at v0.12.0. Every user tracking `:latest` would have stayed on the old release with no signal.

## 2. The iOS job had no upper bound

`Player / iOS` spent 5h49m in `Upload to TestFlight` before the 6-hour default killed it, on a runner billed at 10x. The lane calls `upload_to_testflight(distribute_external: true)`, and external distribution cannot skip waiting for Apple to finish processing, so a slow processing queue parks the job in an unbounded poll loop.

`timeout-minutes: 45` bounds it. The job normally finishes in 17-20 minutes (19m42s on beta.5, 17m8s on beta.2), so this leaves generous headroom. Hitting the timeout fails the publish gate, which is correct: the operator re-dispatches with `allow_missing=ios`, and the IPA reaches TestFlight whenever Apple finishes regardless of what the job reports.

## Verification

`NEWEST` against the live release list:

| query | result |
| --- | --- |
| before | `metadata-relay-v0.9.0` |
| after, stable | `0.12.0` |
| after, prerelease | `0.13.0-beta.6` |

With `NEWEST=0.12.0`, sorting `0.13.0` and `0.12.0` with `sort -V` yields `0.13.0` last, so `IS_NEWEST=true` and the floating tags apply.

- [x] `release.yml` parses (`yaml.safe_load`, 12 jobs); `player-ios.timeout-minutes` reads back as `45`
- [x] Registry state confirming fault 1: `:latest` and `:latest-pg` both resolve to their v0.12.0 digests, `:0.13` does not exist, and `:beta` correctly resolves to `:0.13.0-beta.6`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -330,17 +330,24 @@ jobs:
           # patching an older line (v0.11.2 after v0.12.0) cannot drag
           # :latest, :0 and :0.11 backwards onto the older build. The same
           # applies to :beta against other prereleases.
+          #
+          # Only Mydia's own `vX.Y.Z` tags count. This repository also
+          # publishes the metadata-relay under `metadata-relay-vX.Y.Z`, and
+          # `ltrimstr("v")` leaves that prefix intact, so an unfiltered list
+          # hands `sort -V` a string beginning with a letter. That sorts above
+          # every numeric version, which made v0.13.0 look older than
+          # metadata-relay-v0.9.0 and silently skipped :latest, :0.13 and :0.
           if [ "$IS_PRERELEASE" = "true" ]; then
             CLASS="prerelease"
             NEWEST=$(gh release list --repo "$REPO" --limit 100 \
               --json tagName,isDraft,isPrerelease \
-              --jq '[.[] | select(.isDraft == false and .isPrerelease == true) | .tagName | ltrimstr("v")] | .[]' \
+              --jq '[.[] | select(.isDraft == false and .isPrerelease == true) | .tagName | select(test("^v[0-9]")) | ltrimstr("v")] | .[]' \
               | sort -V | tail -n1)
           else
             CLASS="stable"
             NEWEST=$(gh release list --repo "$REPO" --limit 100 \
               --json tagName,isDraft,isPrerelease \
-              --jq '[.[] | select(.isDraft == false and .isPrerelease == false) | .tagName | ltrimstr("v")] | .[]' \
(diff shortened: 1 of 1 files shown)


arsfeld:
Handled the review point on the tag pattern. `^v[0-9]` is indeed looser than the `vX.Y.Z` the comment claims, so the filter now requires all three numeric components: `^v[0-9]+[.][0-9]+[.][0-9]+`.

Checked against the live release list before changing it: the tighter pattern drops nothing real. Every Mydia release tag still matches, prereleases included (`v0.13.0-beta.6`, `v0.8.1-rc13`), because `test` is not anchored at the end. The set of kept tags is byte-identical between the two patterns today; the tightening only closes the door on a hypothetical future `v<digit>` tag that is not a release, which is exactly the case raised.

Used a `[.]` character class rather than an escaped dot so the expression reads the same through the YAML, shell and jq quoting layers.

Re-verified with the expressions exactly as they now appear in the file: stable `NEWEST` resolves to `0.12.0`, prerelease to `0.13.0-beta.6`.

PR #376, opened 2026-08-09
fix(player): make card taps honest about what they do

arsfeld:
## Problem

Tapping an episode card in the player starts playback immediately. On mobile that is confusing, and there was no way to reach the series details page from a card that plays.

Three surfaces played on tap: the home Continue Watching rail, the home Up Next rail, and the episode rail inside a show's detail page. Everything else already navigated.

Two things made this worse on touch than it sounds. The play affordance on those cards was a **hover** overlay, which never fires on a phone, so identical-looking cards had opposite behavior with nothing to tell them apart. And no card anywhere in the player had an `onLongPress`, so once a tap launched fullscreen playback there was nothing to recover to.

## What Infuse and Plex actually do

The convention both follow comes from the Apple TV app, which states it as two clauses: selecting an item in Continue Watching resumes immediately, and selecting artwork anywhere else opens that item's page.

Plex applies this almost universally. Every poster, card, and episode row opens the preplay screen, which carries Play plus Resume when partially watched. The one documented bypass is Continue Watching, which Plex's own settings docs distinguish from "the Resume button on the details page". Infuse has the same shape and adds a long-press context menu as the escape hatch.

Neither app branches on platform for *where* a tap goes. What changes is how many targets the input can address: Plex Web puts a play button and a "..." menu on hover, and on mobile those collapse into the preplay screen and a long press.

## The rule this implements

> A card's body opens what it depicts. The resume rails are the exception: Continue Watching and Up Next resume on tap. Playback anywhere else comes from a labelled control. Long press opens the paths the tap did not take.

Nothing branches on screen width. `Breakpoints.isDesktop` is width-only (`>= 900px`), so an iPad in landscape reads as desktop; touch affordances here are always present and hover is additive on top.

## Changes

**`MediaCard` now declares what its tap does.** It previously took an opaque `onTap` alongside a doc comment asserting it never started playback, which was false for two of the four rails using it. A required `MediaCardAction` settles it at the call site, and a `play` card draws an always-visible badge instead of relying on hover. The value is derived from the target, so a rail card with nothing playable behind it stops promising a playback its tap cannot deliver.

**A long-press context menu.** New `media_context_menu.dart` offering Play, Go to show, and the episode or movie details screen. Its visibility rules are a pure function over a flattened target, so they are unit-testable without pumping a menu and the rails stay free of item-type knowledge. Wired to long press and to secondary tap, so desktop right-click reaches the same menu. This gives `/episode/:id` a way in from the home screen for the first time, and that screen already links back to its series.

**The show page episode rail picks, the hero plays.** Tapping a card used to resolve a file and push the player route. The hero above it already renders the selected episode with its title, description, and Play button, so the rail now only selects and carries the viewport back to the hero. This costs one extra tap to play the next episode from the show page, which is what Plex charges for the same interaction.

Unchanged: the library, favorites, recently added, unwatched, and collection grids, search results, the home hero, and the player screen. None of them played on tap.

## Testing

Full player suite: 1673 passed, 0 failed. `flutter analyze`: 0 errors, 0 warnings (51 pre-existing info lints, against a baseline of 52 on master).

Worth calling out for reviewers: the `expect(pushed, isEmpty)` assertion in the show-detail rail test is **cheap defense, not the regression guard**. The old code awaited `DeviceContext.detect`, which goes through `connectivity_plus`'s platform channel and never reached the push in a widget test, so that assertion passed against the buggy code too. The discriminating test is the scroll one, which proves `_revealHero` works and could not have passed before this change.

## Still to verify on a device

Widget tests cannot cover either of these:

- Long press versus drag-to-scroll in the horizontal rails. A stationary long press should win the gesture arena, but it needs a real touch device.
- The hero reveal on a phone-sized window: scroll to the episode rail, tap a card, confirm the viewport animates back to a hero describing the tapped episode.

### DIFF
--- a/player/lib/presentation/widgets/media_context_menu.dart
+++ b/player/lib/presentation/widgets/media_context_menu.dart
@@ -47,11 +47,19 @@ class MediaContextTarget {
 /// Pure, so the visibility rules are unit-testable without pumping a menu.
 /// An empty result means the card gets no menu at all.
 List<MediaContextAction> mediaContextActionsFor(MediaContextTarget target) {
+  // A card whose own tap opens the title earns no menu at all. Gating only
+  // `movieDetails` on this left two holes: an episode target got
+  // `episodeDetails` regardless, duplicating what its tap already did, and any
+  // target with a file got `play` even where the rail had no play handler to
+  // run, since `_openMenu` routes Play through a nullable `onItemActivate`.
+  // That is a Play entry that silently does nothing.
+  if (!target.tapPlays) return const [];
+
   return [
     if (target.hasFile) MediaContextAction.play,
     if (target.showId != null) MediaContextAction.goToShow,
     if (target.isEpisode) MediaContextAction.episodeDetails,
-    if (target.isMovie && target.tapPlays) MediaContextAction.movieDetails,
+    if (target.isMovie) MediaContextAction.movieDetails,
   ];
 }
 
--- a/player/lib/presentation/widgets/media_context_menu.dart
+++ b/player/lib/presentation/widgets/media_context_menu.dart
@@ -1,0 +1,139 @@
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+
+import '../../core/theme/colors.dart';
+
+/// What a card's long-press menu can offer.
+enum MediaContextAction { play, goToShow, episodeDetails, movieDetails }
+
+/// The subject of a long-press menu, flattened out of whichever rail item the
+/// card was built from.
+///
+/// The menu deliberately knows nothing about `ContinueWatchingItem`,
+/// `UpNextItem` or `RecentlyAddedItem`. Each converts itself into one of these
+/// where the card is built, so a new rail item type needs no change here.
+class MediaContextTarget {
+  /// What the card depicts: a movie id or an episode id.
+  final String id;
+
+  /// 'movie', 'tv_show', or 'episode'.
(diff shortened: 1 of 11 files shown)


player/lib/presentation/widgets/media_context_menu.dart
arsfeld, line 1: Good catch, and the hole was slightly wider than described.

Gating only `movieDetails` on `tapPlays` left two problems. An episode target got `episodeDetails` regardless of `tapPlays`, duplicating what its own tap already did. And any target carrying a file got `play` even when the rail had no play handler: `_openMenu` routes Play through `widget.onItemActivate?.call(item)`, which is nullable, so on a rail without an activate handler that entry would have opened a menu and then silently done nothing.

Not reachable from the call sites in this PR. Both play rails pass `onItemActivate`, and the "Similar in your library" rail carries `RecentlyAddedItem`, which produces a target with `hasFile: false` and no `showId`, so its menu was already empty. But `ContentRail` is a reusable widget and this was a live trap for the next rail added to it.

Fixed in 866dd594 with an early return, so a card whose tap opens the title earns no menu at all, which is what the `tapPlays` doc comment claimed all along:

```dart
if (!target.tapPlays) return const [];
```

Added two regression tests: a playable episode with `tapPlays: false` earns an empty menu, and no target that navigates on tap is ever offered Play across all three type values. All six existing tests still pass unchanged.

PR #380, opened 2026-08-09
fix(player): make stuck downloads recoverable instead of delete-and-re-add

arsfeld:
## Problem

A download in the player could get stuck showing "Downloading" with a frozen progress bar, and the only way out was deleting it and adding it again. There was no restart, no resume, and nothing that noticed a download had stopped making progress.

On iOS this is not an edge case. The OS suspends the app shortly after backgrounding, the Dart isolate stops, the sockets die, and the Hive row still says `downloading`. Nothing on the next launch picked it back up.

## Root causes, verified

Three independent defects, each confirmed by a regression test that fails against `master`:

**Orphaned tasks were never reaped.** Liveness lives in `_cancelTokens`, an in-memory map that is empty after a relaunch. Only `transcoding` rows were resumed, by `_resumeTranscodingTasks`. Rows in `downloading` and `pending` sat forever.

**Nothing could restart a task in a non-terminal state.** `retryDownload` returned early unless the status was `failed` or `cancelled`; `resumeDownload` required exactly `paused`. A task frozen in `downloading`, `transcoding`, `pending`, or `queued` matched neither, so no API could move it. The UI agreed: the active-download sheet offered only Cancel.

**Resume restarted from zero and corrupted the file.** Both download paths called `_generateFileName`, which stamps `DateTime.now().millisecondsSinceEpoch` into the name, so the `if (await file.exists())` resume check tested a path that had never existed. Separately, `Dio.download` defaults to `FileAccessMode.write`, which truncates, so a `Range: bytes=N-` request wrote the tail of the file at offset zero. The tests captured all three symptoms before the fix:

```
Expected: '/tmp/.../partial.mp4'
  Actual: '/tmp/.../Seeded_1080p_1786250113314.mp4'

Expected: 'bytes=4-'
  Actual: <null>

PathNotFoundException: Cannot open file, path = '/tmp/.../partial.mp4'
```

Retry loops were also unbounded: a non-cancel `DioException` during an incomplete transcode slept two seconds and continued forever, so a dead server produced an infinite spin rather than a failure anyone could act on.

## What changed

**State.** Two new non-terminal statuses, `interrupted` (was active, no loop driving it, partial file intact) and `stalled` (has a loop, bytes stopped moving). Two new `DownloadTask` fields as Hive 33 and 34, `lastProgressAt` and `recoveryAttempts`, both appended with defaults so boxes written by older builds still deserialize.

**Recovery.** A pure `planRecovery` function decides what to do with every task that claims to be active but has no live token. It runs at launch, when the job service is injected, and on app resume through the observer already in `app.dart`. It honours `maxConcurrentDownloads` and `autoStartQueued`, and gives up into `failed` after three attempts rather than looping.

**Resume.** Both paths now reuse the recorded `filePath`, take the offset from the actual on-disk length rather than a persisted counter that can disagree after a crash, and pass `FileAccessMode.append` for ranged requests.

**Watchdog.** A 30-second tick marks a `downloading` task stalled after 90 seconds without byte movement, and a `transcoding` task after 5 minutes, then re-drives it.

**API and UI.** New `restartDownload` accepting any status but `completed`; `resumeDownload` widened to `interrupted` and `stalled`; `retryDownload` reduced to a thin call into restart. The active sheet gains Pause, Resume, and Restart. Interrupted and stalled cards show what is happening and offer inline Resume instead of a frozen progress bar that implies work.

**Errors.** Transient failures now retry three times with 2s, 4s, 8s backoff and end as `interrupted`. A 404 from `getJobStatus` means the server dropped the job, so the task fails with a message saying so rather than retrying against something that no longer exists.

## A bug the old tests were hiding

`download_service_integration_test.dart` was 1359 lines, roughly 620 of which were `TestDownloadService` and `TestDownloadDatabase`. Neither implemented its interface, and `TestDownloadService` shared no code with `_NativeDownloadService`, so its twenty-odd tests asserted the behaviour of an implementation that ships to nobody. That is worse than no tests, because the suite reads as covered.

Re-pointing those tests at the real service through a shared harness immediately surfaced a genuine bug: **pausing a live download silently cancelled it.** `pauseDownload` wrote `paused`, then the `DioExceptionType.cancel` handler in `_startDownloadTask` unconditionally overwrote it with `cancelled`.

```
Expected: 'paused'
  Actual: 'cancelled'
```

The cancel handler now only claims `cancelled` for a task still marked `downloading`, and `pauseDownload` claims the status before cancelling the token. `restartDownload` was exposed to the same race and is covered by the same guard. This mattered immediately, because the Pause control added in this branch would otherwise have cancelled downloads.

Also fixed along the way: the concurrency setting in the storage sheet was written to Hive and never reached the downloader, since `setMaxConcurrentDownloads` had no callers. And `DownloadQueueManager.processQueue` was dead code that never worked, setting a task to `pending` and then calling `retryDownload`, which returns early on `pending`.

## Verification

- 1772 tests pass, no skips. `dart analyze --fatal-warnings` is clean, and the info count dropped from 52 to 50 with the fake service deleted.
- The integration test file went from 1359 lines to 705, now exercising the code that ships.
- Test counts were checked across repeated runs to rule out the intermittent stream-close failure seen mid-development.

**Not verified:** the two manual iOS checks in the plan, backgrounding for two minutes and force-quitting mid-download, have not been run. There is no simulator in the environment this was built in. They are the last thing worth doing before merge.

## Out of scope

Downloads that continue while the app is suspended. That needs migrating off dio onto iOS background `URLSession`, most likely via `background_downloader`, and it changes the progressive-transcode UX because `URLSession` wants one URL producing one file rather than a file that grows while being read. The recovery planner, watchdog, and UI controls here all remain useful underneath such an engine.


arsfeld:
Thanks, three of these were real and are fixed in bc60fb6. One is not, and I want to record why rather than change working code.

## Not applying: the `_applyRecovery` switch fall-through

The claim is that cases fall through and that Dart requires each `case` to end in `break`/`return`. That was true in Dart 2, but Dart 3 changed it: switch statement cases no longer fall through and no longer need a terminator. `dart analyze --fatal-warnings` is clean on this file, and it would reject a non-terminated case under the old rule.

The behaviour is also pinned by tests that would fail if fall-through were happening:

- `fails a task that has exhausted its recovery attempts` asserts the task ends `failed`. Falling through to `requeue` would leave it `queued`.
- `parks every orphan as queued when autoStart is off` asserts all three orphans end `queued`. Falling through to `resume` would leave them `interrupted`.

Both pass. Adding `return` statements would be harmless but would imply a hazard that does not exist here.

## Fixed: `downloadManager` rebuild disposing the service

This one was the most serious thing in the review and it was a regression in this branch. `downloadManager` awaited `ref.watch(downloadSettingsProvider.future)`, so saving the storage sheet invalidated the settings provider, rebuilt the manager, and ran its `ref.onDispose` teardown, cancelling every download in flight. Changing the concurrency limit would have killed your active downloads.

Settings are now read once and applied through `ref.listen`. There is a regression test, and I verified it has teeth by restoring the `watch` and confirming it fails:

```
Expected: true
  Actual: <false>
```

## Fixed: `transcoding` counted as both active and queued

Correct, and it predates this branch. A transcoding task appeared in `activeDownloads` and `queuedDownloads`, so `totalPending` double-counted it. It also made the "Cancel N queued downloads" button overstate itself, because `cancelAllQueued` only acts on `queued` and `pending`. `transcoding` is now active only.

## Also fixed while in here

Every progress emission now routes through a guarded `_emit`. Downloads are fire-and-forget, so a loop could outlive `dispose` and throw `Cannot add new events after calling close`. That had been showing up as an intermittent test failure, and it would be a real crash on teardown. Three consecutive runs of the download suite are now clean at 101 tests.

Full suite: 1774 passing, `dart analyze --fatal-warnings` clean.

PR #395, opened 2026-08-10
feat(relay): GitHub SSO for the maintainer dashboards, and file issues from feedback

arsfeld:
Puts the metadata-relay maintainer dashboards behind GitHub App sign-in, and lets you file a GitHub issue straight from a feedback submission as yourself.

Following the deep link in a feedback notification email currently means an HTTP Basic Auth prompt, which is miserable on a phone. More importantly, basic auth carries no identity, so there was no way to act on GitHub from the dashboard without parking a long-lived write token in the relay's environment.

## What changed

`/feedback` and `/errors` authenticate against a GitHub App ("Mydia Relay") using user-to-server authorization, gated on membership of the `getmydia` organization. The same token files the issue, so there is one credential pair and no permanent write token on the server. A GitHub App rather than a classic OAuth app keeps the token narrow: it can only touch Issues on repos the App is installed on, instead of every public repo the account owns.

The feedback dashboard gains a "File GitHub issue" button that opens a modal with an editable title and body, then records the issue number and URL back onto the submission and moves it to a new `filed` state.

## Two modes, mutually exclusive

Setting `DASHBOARD_GITHUB_ORG` switches the dashboards to GitHub sign-in and turns basic auth off. Leaving it empty keeps today's behaviour exactly, which is what self-hosted relays and the test suite use. `test/metadata_relay_web/dashboard_auth_test.exs` is byte-identical to master after this branch, as the proof of that.

The issue-filing button only appears for a signed-in maintainer, so a relay in basic-auth mode simply does not show it.

## Security fix worth calling out separately

`config/config.exs` shipped a hardcoded placeholder `secret_key_base` with no runtime override. Harmless while the session carried nothing, but forgeable by anyone reading the repo once a login session exists. `SECRET_KEY_BASE` is now read from the environment, and the session cookie is encrypted rather than only signed.

When the variable is absent the relay generates a random key at boot and warns, rather than refusing to start. Raising would have broken every existing self-hosted relay on upgrade for a variable that did not previously exist.

## Compatibility

No public API surface moves. `lib/metadata_relay/router.ex` is unchanged, so the TMDB, TVDB and subtitle proxy paths, feedback ingest, pairing claims, and the p2p access endpoints all behave exactly as before. No Mydia instance is affected by this deploy.

Merging deploys nothing on its own: the relay ships only on a `metadata-relay-v*` tag, so this needs `metadata-relay-v0.12.0` cut against master afterwards.

Order does matter for the cluster config, and the ordering constraint is one way only. The ConfigMap in this branch sets `DASHBOARD_GITHUB_USERS: "arsfeld"`, and a non-empty allowlist turns basic auth off. Applying that ConfigMap before `GITHUB_APP_CLIENT_ID` and `GITHUB_APP_CLIENT_SECRET` are in the secret leaves `/feedback` and `/errors` unreachable: basic auth is disabled and sign-in cannot complete. The relay itself is unaffected and keeps serving the proxy, and adding the secret restores access, so the failure is a locked door rather than an outage. There is no window in which the dashboards are exposed, only one in which they are shut. Apply the secret first, or ship the ConfigMap with the allowlist empty and fill it in once the App exists.

## Operator steps before this does anything

1. Create a GitHub App named "Mydia Relay" owned by `getmydia`. Repository permissions: Issues, read and write. Callback URL `https://relay.mydia.dev/auth/github/callback`. Leave "Expire user authorization tokens" unchecked, since there is deliberately no refresh flow. Disable webhooks.
2. Install it on `getmydia/mydia`.
3. Add `SECRET_KEY_BASE`, `GITHUB_APP_CLIENT_ID` and `GITHUB_APP_CLIENT_SECRET` to the cluster secret, and `DASHBOARD_GITHUB_USERS` plus `FEEDBACK_GITHUB_REPO` to the ConfigMap.

## Testing

315 tests, 0 failures under `mix test --warnings-as-errors`. New coverage includes forged and missing OAuth `state`, off-site `return_to` rewriting, logins outside the allowlist, a login removed from the allowlist after signing in, the full GitHub client error surface, and an explicit assertion that `contact`, `source_ip` and `instance_id` never reach an issue body headed for a public repo.

Issues are public, so the draft omits the reporter's contact address, source IP, and instance ID by default, and the modal exists as a review step before anything is published.

## Summary by CodeRabbit

* **New Features**
  * Added GitHub sign-in with active organization membership checks for dashboard access; basic authentication remains available as a fallback.
  * Added tools to draft and file feedback as GitHub issues directly from the dashboard.
  * Added filed-status tracking, issue links, filtering, focused submission links, and improved issue review controls.
  * Enhanced encrypted session handling and access verification during GitHub outages.

* **Documentation**
  * Documented dashboard authentication modes, GitHub setup, repository configuration, and required secrets.

Rebased onto master on 19 August 2026, picking up the SubDL subtitle work and the Trakt removal that landed in the meantime. Three files needed hand resolution, all documentation: the relay `CLAUDE.md` env var list, the infra README, and `secret.yaml.example`, each of which had gained `SUBDL_API_KEY` on master and lost the OpenSubtitles keys. `config/runtime.exs` and the feedback notifier merged cleanly, and `dashboard_auth_test.exs` is still byte-identical to master.

One bug surfaced while re-reviewing after the rebase, fixed in `fix(relay): deny malformed OAuth callback params instead of raising`. Nested query params arrive as lists, and both `Plug.Crypto.secure_compare/2` and `OAuth.exchange_code/1` only accept binaries, so `/auth/github/callback?code=x&state[]=y` raised `FunctionClauseError`. Not an auth bypass, since the request was refused either way, but an unauthenticated caller could turn every such request into a 500 and an ErrorTracker entry. The callback head now guards on binary `code` and `state`, so malformed params fall through to the existing catch-all and redirect to the login page like any other failed sign-in.

`infra/deploy` rebuilds `metadata-relay-secrets` from `metadata_relay_secret_data` on every run, and it did not know about the new keys, so credentials added by hand with `kubectl` would have been dropped on its next run. `chore(infra): provision the dashboard sign-in secrets from infra/deploy` adds them as optional config fields written only when set, matching how `SUBDL_API_KEY` and `SMTP_PASSWORD` already work. The Phoenix session key gets its own field rather than reusing `relay_secret_key_base`, which despite the name is written as `RELAY_TOKEN_SECRET`.

Worth knowing separately, and not something this branch changes: `DASHBOARD_USERNAME` and `DASHBOARD_PASSWORD` appear nowhere in `infra/deploy`, so they are only in the live secret by hand. A run of `infra/deploy` drops them, and in basic-auth mode `runtime.exs` raises on a missing username in prod, which would crashloop the relay. That is why this branch does not make GitHub mode fall back to basic auth when the App credentials are absent: the fallback could take the proxy down instead of just the dashboards.

## Access follows org membership, not a list

Updated after review: `DASHBOARD_GITHUB_USERS` is gone, replaced by `DASHBOARD_GITHUB_ORG=getmydia`. Anyone with an active membership of that organization gets in, so adding a maintainer to the org grants the dashboards and removing them takes it away, with no list of logins to keep in step in the ConfigMap.

Membership is read with `GET /user/memberships/orgs/{org}` using the signed-in maintainer's own token. That endpoint is scoped to the caller, so it sees private membership and the App needs no permission to read the org's member list. Only `active` counts. A pending invitation also answers 200, and accepting that would mean merely inviting someone handed them the dashboard, so there is an explicit test for it.

The real cost of this over an allowlist is revocation latency, and it is worth being honest about. The allowlist was config, so it could be consulted on every request and a removal took effect immediately. Membership is a remote lookup, so it is checked at sign-in and re-checked at most every five minutes. Someone removed from the organization keeps a live session until that window passes.

The other direction matters too: an unreachable GitHub leaves an already-verified session alone rather than signing it out. Only a definitive refusal, a 404 or a non-active state or a 401/403, ends the session. Otherwise a GitHub outage would lock maintainers out of the dashboards they need precisely when something is going wrong.

`dashboard_auth_test.exs` is still byte-identical to master, so basic-auth mode remains provably untouched.

326 tests, 0 failures. `runtime.exs` has no test coverage, so the mode switch was exercised directly: unset, empty string, and whitespace all fall back to basic auth, and a set value produces GitHub mode with basic-auth credentials left nil.

### DIFF
--- a/infra/config.yaml.example
+++ b/infra/config.yaml.example
@@ -29,6 +29,18 @@ relay_tmdb_api_key: "your-tmdb-api-key"
 # Key from: https://subdl.com/panel/api
 relay_subdl_api_key: ""
 
+# Maintainer dashboard GitHub sign-in (optional)
+# Only used when DASHBOARD_GITHUB_USERS is set in the metadata-relay ConfigMap.
+# Leave blank and the dashboards stay on basic auth.
+# The session key is the Phoenix one, unrelated to relay_secret_key_base above,
+# which is written as RELAY_TOKEN_SECRET. Generate with:
+#   head -c 48 /dev/urandom | base64
+# Credentials come from the "Mydia Relay" GitHub App:
+#   https://github.com/organizations/getmydia/settings/apps
+relay_dashboard_secret_key_base: ""
+relay_github_app_client_id: ""
+relay_github_app_client_secret: ""
+
 # Feedback email notifications (optional)
 # All six fields below must be set together for the relay to send feedback emails.
 # Leave blank to disable. Stored locally only — never committed to git.
--- a/infra/config.yaml.example
+++ b/infra/config.yaml.example
@@ -30,7 +30,7 @@ relay_tmdb_api_key: "your-tmdb-api-key"
 relay_subdl_api_key: ""
 
 # Maintainer dashboard GitHub sign-in (optional)
-# Only used when DASHBOARD_GITHUB_USERS is set in the metadata-relay ConfigMap.
+# Only used when DASHBOARD_GITHUB_ORG is set in the metadata-relay ConfigMap.
 # Leave blank and the dashboards stay on basic auth.
 # The session key is the Phoenix one, unrelated to relay_secret_key_base above,
 # which is written as RELAY_TOKEN_SECRET. Generate with:
--- a/infra/deploy
+++ b/infra/deploy
@@ -100,6 +100,13 @@ class Config:
     # relay answers /api/v1/subtitles/* with 503 not-configured)
     relay_subdl_api_key: str = ""
 
+    # Maintainer dashboard sign-in (optional; leave blank and the dashboards
+    # stay on basic auth). Note this is the Phoenix session key, unrelated to
+    # relay_secret_key_base above, which is written as RELAY_TOKEN_SECRET.
+    relay_dashboard_secret_key_base: str = ""
+    relay_github_app_client_id: str = ""
+    relay_github_app_client_secret: str = ""
+
     # Feedback email notifications (optional; leave blank to disable)
     relay_smtp_host: str = ""
     relay_smtp_port: str = ""
@@ -193,6 +200,9 @@ class Config:
             "relay_tmdb_api_key": self.relay_tmdb_api_key,
             "relay_smtp_password": self.relay_smtp_password,
             "relay_subdl_api_key": self.relay_subdl_api_key,
+            "relay_dashboard_secret_key_base": self.relay_dashboard_secret_key_base,
+            "relay_github_app_client_id": self.relay_github_app_client_id,
+            "relay_github_app_client_secret": self.relay_github_app_client_secret,
             "relay_smtp_host": self.relay_smtp_host,
             "relay_smtp_port": self.relay_smtp_port,
             "relay_smtp_username": self.relay_smtp_username,
@@ -1036,6 +1046,15 @@ def metadata_relay_secret_data(config: Config) -> dict[str, str]:
 
     if config.relay_subdl_api_key:
         data["SUBDL_API_KEY"] = config.relay_subdl_api_key
+
+    if config.relay_dashboard_secret_key_base:
+        data["SECRET_KEY_BASE"] = config.relay_dashboard_secret_key_base
+
+    if config.relay_github_app_client_id:
+        data["GITHUB_APP_CLIENT_ID"] = config.relay_github_app_client_id
+
+    if config.relay_github_app_client_secret:
+        data["GITHUB_APP_CLIENT_SECRET"] = config.relay_github_app_client_secret
 
     return data
 
--- a/infra/kubernetes/apps/metadata-relay/README.md
+++ b/infra/kubernetes/apps/metadata-relay/README.md
@@ -75,7 +75,7 @@ curl https://relay.mydia.dev/health
 - `MIX_ENV`: Environment (prod)
 - `SQLITE_DB_PATH`: Path to SQLite database file
 - `FEEDBACK_DASHBOARD_URL`: Base URL used in feedback notification links
-- `DASHBOARD_GITHUB_USERS`: Comma-separated GitHub logins allowed into `/feedback` and `/errors`. Setting this switches the dashboards from basic auth to GitHub sign-in. Leave empty to keep basic auth.
+- `DASHBOARD_GITHUB_ORG`: GitHub organization whose active members may reach `/feedback` and `/errors`. Setting this switches the dashboards from basic auth to GitHub sign-in. Leave empty to keep basic auth.
 - `FEEDBACK_GITHUB_REPO`: Repository that issues filed from the feedback dashboard land in (default `getmydia/mydia`)
 
 ### Secrets (secret.yaml)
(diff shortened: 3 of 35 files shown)


arsfeld:
All three review comments were valid and are fixed in 60187a48.

**`runtime.exs` requiring basic-auth credentials in GitHub-only mode.** This was the substantive one: the code contradicted the documentation in the same PR. `DASHBOARD_USERNAME` and `DASHBOARD_PASSWORD` are now only required when `DASHBOARD_GITHUB_USERS` is empty, so a GitHub-only deployment no longer has to invent unused credentials. A deployment with neither configured still fails fast, and the message now names both options rather than only the basic-auth one.

Verified by evaluating the config as prod both ways: with the allowlist set and no basic credentials it resolves cleanly, and with neither set it raises `DASHBOARD_USERNAME not set, and DASHBOARD_GITHUB_USERS is empty`.

**OAuth `state` left in the session.** Fixed. `sign_in/3` now deletes both `oauth_state` and `return_to`, making the state single-use so a replayed callback URL cannot be validated against a stale value. Covered by a new test asserting both keys are gone after a successful callback.

**`submit_issue` assuming a draft and token are present.** Fixed, with one correction to the diagnosis: a nil draft raises in `handle_event` itself, before `start_async`, so that path crashed the LiveView rather than producing the transport error. The nil-token path is where the misleading "GitHub is unreachable" came from, since `create_issue/2` has an `is_binary(token)` guard and the `FunctionClauseError` surfaced as an async exit. Both are now handled explicitly: a missing draft reports "No issue draft is open" and a missing token asks the maintainer to sign in again. Two new tests cover them, including an assertion that the token case does *not* say "unreachable".

Suite is 240 tests, 0 failures.

arsfeld:
Thanks, four of the five were worth acting on. Summary of what happened to each.

**Empty dashboard credentials (`runtime.exs`) — fixed, and it was live.** Confirmed exploitable before fixing: an empty string is truthy in Elixir, so `System.get_env("DASHBOARD_USERNAME") || raise(...)` yields `""`, and `Plug.BasicAuth` (1.18.1, the version this project locks) then accepts `Authorization: Basic <base64(":")>`. Both variables now go through `normalize_env`. Prod raises on a blank value, dev falls back to the admin default. This predates the branch but the same lines were being rewritten here.

**Both GitHub App credentials required in GitHub mode — addressed, but warning rather than raising.** The concern is real and the dashboards are genuinely unreachable in that state. Raising at boot would crashloop the relay and take the TMDB/TVDB proxy down for every Mydia install over a maintainer-only feature. The misconfiguration also fails closed, not open, so the dashboards are shut rather than exposed. Boot now prints a specific warning naming both variables and the consequence.

**Mentions and issue references in the issue body — fixed.** Good catch, and it is more than cosmetic: feedback is submitted anonymously by any install, so a reporter could make a maintainer's account ping arbitrary people or backlink an unrelated issue. Both are escaped with an empty HTML comment, so the text reads identically to a human. The escape is deliberately narrow, leaving `a@b.com`, `C# code` and `100% broken` untouched, with tests both ways.

**Duplicate issues after a partial filing failure — acknowledged, not fixed here.** The window is real but narrow: a database write that fails immediately after a successful `POST /issues`. The LiveView already reports it distinctly ("The issue was created but could not be recorded. Reload and check GitHub.") rather than as a transport error, so a maintainer is told rather than silently retrying. A reservation state machine plus a recovery path is disproportionate to a duplicate public issue that a human can close, and it would land unreviewed at the end of an already large branch. Worth its own issue.

**Requiring `FEEDBACK_GITHUB_REPO` in production — not changing.** `getmydia/mydia` is the correct repository for this deployment, which is the only one running GitHub mode. A self-hosted relay that set the org but forgot the repo would fail on the GitHub side anyway, since its App token has no access to `getmydia/mydia`, so the fallback surfaces as a visible error in the modal rather than a stray public issue.

Note that the allowlist these comments were written against is gone. `DASHBOARD_GITHUB_USERS` has been replaced by `DASHBOARD_GITHUB_ORG`, and access now follows active membership of that organization.

PR #400, opened 2026-08-11
fix(downloads): stop auto-rejecting healthy qBittorrent downloads in a loop

arsfeld:
## The bug

Manually grabbing a 720p or 1080p release starts the download, then Mydia cancels it a minute or two later, blacklists the release, and grabs a different one. That replacement meets the same fate, and so on until the result set is exhausted. With the quality profile on "Any" the replacement is a 4K release; with the profile locked to HD-720 it walks the 720p list and then stops with nothing downloaded.

Reported against "Minions & Monsters (2026) [720p] [WEBRip]", which cycled through the WEBRip, the LAMA release, and the YTS release.

## Root cause

The pre-completion content check added in 81881830 and shipped in v0.13.0 reads `DownloadStatus.files` and assumes it is a list of leaf files. For qBittorrent it is not: `qbittorrent.ex` populates it with a single `content_path`, which the qBittorrent API defines as *"root path for multifile torrents, absolute file path for singlefile torrents"*.

Nearly every release is packaged as a folder, so the check ran `Path.extname/1` over a directory name:

- `Minions.Monsters.2026.720p.WEBRip.x264-YTS` yields `.x264-YTS`
- `Minions & Monsters (2026) [720p] [WEBRip]` yields `""`

Neither is a video extension, so a healthy torrent was judged content-free and handed to `Queue.reject_release/2`, which blacklists the release, removes the torrent with `delete_files: true`, and queues a replacement search.

rtorrent had the identical bug via `base_path`. Transmission and rqbit were unaffected: both build `files` from genuine per-file listings.

## The fix

- **`Client.list_files/3`**, a new optional adapter callback. Adapters that cannot enumerate a torrent's contents simply do not implement it and report `:unsupported` via `@optional_callbacks` plus a `function_exported?/3` probe, so they are safe by construction rather than by review.
- **qBittorrent enumerates for real** via `/api/v2/torrents/files`. That endpoint also reports names without the `.!qB` suffix, which fixes a second false positive for anyone running "Append .!qB extension to incomplete files".
- **Transmission and rqbit** implement the callback by delegating to the leaf listings they already build.
- **The check is gated on a real enumeration.** `{:error, _}` and `{:ok, []}` both mean "unknown" and never reject; an empty list from a torrent client means metadata has not resolved yet.
- **`content_checked_at`** bounds the cost to one client request per download for its lifetime, rather than one per poll.
- **A cap of 3 consecutive auto-rejections per media item**, reusing `search_backoffs`, bounds any future false positive to three releases instead of the entire result set. When it trips the download is left completely alone to finish and import, since the likely truth is that the detector is wrong.
- **Honest failure reasons.** `reject_release/2` no longer hardcodes `rejected_by_user` for system rejections.
- **A migration purges the damage**: `rejected_by_user` rows written on or after the v0.13.0 tag date. Genuine operator rejections from that four-day window go too, which is the accepted trade so every affected install self-heals without the operator having to find an admin page.

Malware protection is preserved throughout. The existing regression test for the disguised `.exe` in a House of the Dragon torrent still passes, and qBittorrent now gets that protection for the first time.

rtorrent keeps no pre-completion protection until `f.multicall` is wired up, but it is now safe rather than destructive.

## Verification

- Full SQLite suite: 6857 tests, 0 failures, 34 skipped
- PostgreSQL feature suites green. The 4 failures in the full Postgres run are the known deep-worktree-path flake: they die in `create_test_library_path/3` at fixture setup with a 256-character `:tmp_dir` against a `varchar(255)` column, before any changed code runs.
- `mix credo --strict`: no issues
- `mix format --check-formatted`: clean

### DIFF
--- a/lib/mydia/config/schema.ex
+++ b/lib/mydia/config/schema.ex
@@ -97,6 +97,11 @@ defmodule Mydia.Config.Schema do
       # inserted without an explicit `expires_at`. See `Mydia.Downloads.Blacklists`
       # (#123).
       field :release_blacklist_default_ttl_days, :integer, default: 30
+      # How many consecutive system auto-rejections (see
+      # Mydia.Jobs.DownloadMonitor) one media item may accrue before Mydia
+      # stops auto-rejecting it. Bounds a systematic false positive to this
+      # many releases instead of the item's entire search result set.
+      field :auto_reject_limit, :integer, default: 3
     end
 
     embeds_one :upgrades, Upgrades, on_replace: :update, primary_key: false do
@@ -350,9 +355,14 @@ defmodule Mydia.Config.Schema do
 
   defp downloads_changeset(schema, attrs) do
     schema
-    |> cast(attrs, [:monitor_interval_minutes, :release_blacklist_default_ttl_days])
+    |> cast(attrs, [
+      :monitor_interval_minutes,
+      :release_blacklist_default_ttl_days,
+      :auto_reject_limit
+    ])
     |> validate_number(:monitor_interval_minutes, greater_than: 0)
     |> validate_number(:release_blacklist_default_ttl_days, greater_than: 0)
+    |> validate_number(:auto_reject_limit, greater_than: 0)
   end
 
   defp upgrades_changeset(schema, attrs) do
--- a/lib/mydia/downloads.ex
+++ b/lib/mydia/downloads.ex
@@ -129,6 +129,8 @@ defmodule Mydia.Downloads do
   @spec update_download(Download.t(), map()) :: {:ok, Download.t()} | {:error, Ecto.Changeset.t()}
   defdelegate update_download(download, attrs), to: Mydia.Downloads.History
 
+  defdelegate mark_content_checked(ids), to: Mydia.Downloads.History
+
   @doc """
   Lists failed path-mapping-mismatch downloads whose reported path is at or under
   the given remote prefix (the apply-mapping fan-out set).
--- a/lib/mydia/downloads/client.ex
+++ b/lib/mydia/downloads/client.ex
@@ -302,6 +302,25 @@ defmodule Mydia.Downloads.Client do
   """
   @callback supported_protocols() :: [protocol(), ...]
 
+  @doc """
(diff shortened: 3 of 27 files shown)


arsfeld:
Thanks, all five are addressed in 58444391. Three were real, two were comment accuracy.

**1. Counter incremented before the rejection succeeded** — correct, and a genuine bug. `record_failure/3` now runs only inside the `{:ok, :rejected}` branch. As written before, a release with no indexer or guid would fail `Blacklists.extract_key/1` every pass, increment the counter anyway, and after three passes permanently suppress real auto-rejections for that media item without ever having rejected anything.

**2. Blank `save_path` yielding relative paths** — correct. `save_path_from_info/1` now returns a parse error when `save_path` is missing or blank rather than defaulting to `""`. That fails closed: the caller treats the torrent as unevaluated and leaves it alone, which is the safe direction, instead of receiving relative paths that claim to be absolute. Covered by a new test.

**3. The regression test did not reproduce the failure mode** — the best catch of the five, and entirely right. That test stubbed Transmission with a leaf `.mkv`, so it would have passed before the fix too and was worthless as regression coverage. Replaced with two tests driving a real qBittorrent adapter whose `/torrents/info` reports `content_path` as the release folder (`/downloads/Minions.Monsters.2026.720p.WEBRip.x264-YTS`, whose `Path.extname/1` is `.x264-YTS`):

- one asserting the download survives when the real enumeration holds an `.mkv`, which is the exact shape that caused the loop
- one asserting a torrent whose enumeration holds only `payload.exe` is still rejected, so gating on enumeration does not cost qBittorrent users the malware protection this check exists for

**4. rqbit comment describing Transmission's `torrent-get`** — correct, that was a copy-paste from the Transmission implementation. Rewritten to describe rqbit's own details endpoint, including the `output_folder` join and the `included: false` skip.

**5. Migration comment calling the literal ISO-8601** — correct, ISO-8601 uses a `T` separator and the literal is space-separated. Reworded to "timestamp literal matching the stored format", with the reason spelled out: SQLite compares it lexicographically as text, so a format differing from what is stored would compare wrong.

PR #401, opened 2026-08-11
fix(metadata-relay): answer 503 when Trakt credentials are absent

arsfeld:
## What was broken

Connecting a Trakt account fails on every Mydia instance with:

> Failed to start Trakt authorization. Please try again.

Reported against the galactica deployment, but it is not deployment-specific. `user_integrations` there has zero rows: Trakt has never successfully connected for anyone since it shipped in f47d58c9 (2026-02-22).

## Root cause

Traced through each boundary:

| Layer | Result |
| --- | --- |
| LiveView `trakt_connect` → `TraktClient.generate_device_code/0` | ok |
| `POST relay.mydia.dev/trakt/oauth/device/code` | request arrives |
| relay router → `TraktHandler.generate_device_code/0` | dispatches |
| `MetadataRelay.Trakt.Client.client_id/0` | **raises** |

From the live relay pod:

```
** (RuntimeError) TRAKT_CLIENT_ID environment variable is not set.
    lib/metadata_relay/trakt/client.ex:86: MetadataRelay.Trakt.Client.client_id/0
    lib/metadata_relay/trakt/handler.ex:20: MetadataRelay.Trakt.Handler.generate_device_code/0
    lib/metadata_relay/router.ex:1251: MetadataRelay.Router.handle_trakt_request/2
```

`handle_trakt_request/2` only handles `{:error, _}` tuples, so the raise escaped to Plug's generic HTML 500. Mydia could not distinguish that from a transient outage and told users to retry something that can never succeed.

The credentials were never recorded anywhere either: not in `secret.yaml.example`, `configmap.yaml`, the README env table, `docker-compose.yml`, or the live cluster secret. There were also no tests covering any Trakt route.

## What this changes

- `Client.configured?/0`, with both Trakt router helpers guarded by it. An unconfigured relay answers `503 {"error": "trakt_not_configured"}` rather than raising. A relay without Trakt credentials still serves TMDB and TVDB normally.
- Blank credentials count as missing. Only `nil` and `false` are falsy in Elixir, so an empty `TRAKT_CLIENT_ID` previously flowed through as an empty `trakt-api-key` header and drew a 403 from Trakt. Second latent bug in the same path.
- The integrations LiveView matches the relay's error code and names what actually has to change instead of saying "try again".
- `TRAKT_CLIENT_ID` / `TRAKT_CLIENT_SECRET` documented alongside the other provider credentials.
- 10 tests over the unconfigured path, where there were none.

## Verification

- Reproduced the 500 by curling `relay.mydia.dev` directly, with Mydia out of the picture.
- The new tests were confirmed to fail against pre-fix code with the exact production stack trace (`client.ex:86`), then pass after.
- metadata-relay: 185 tests, 0 failures. Mydia integrations LiveView: 5 tests, 0 failures. Both projects format-clean.

## Still needed to actually enable Trakt

This makes the misconfiguration diagnosable; it does not provision credentials. To turn Trakt on, register an app at https://trakt.tv/oauth/applications (redirect URI `urn:ietf:wg:oauth:2.0:oob` for the device flow) and add both keys to the `metadata-relay-secrets` secret on k3s-can1.

No mydia-side Bypass test: `METADATA_RELAY_URL` is read from global env, and the repo already documents that mutating it races async tests (`test/mydia/media/refresh_test.exs:144`). The relay tests carry the contract instead.

## Cross-version compatibility

Both directions degrade to the existing generic message, so relay and app can roll independently: an old Mydia against the new relay gets a 503 it does not specially match; a new Mydia against an old relay gets the HTML 500 it does not match either.

### DIFF
--- a/infra/kubernetes/apps/metadata-relay/secret.yaml.example
+++ b/infra/kubernetes/apps/metadata-relay/secret.yaml.example
@@ -16,6 +16,14 @@ stringData:
   TMDB_API_KEY: "your-tmdb-api-key"
   TVDB_API_KEY: "your-tvdb-api-key"
 
+  # REQUIRED for the Trakt.tv integration (scrobbling and sync).
+  # Create an application at https://trakt.tv/oauth/applications with the
+  # redirect URI urn:ietf:wg:oauth:2.0:oob (device flow).
+  # Without these, every /trakt/* endpoint answers 503 trakt_not_configured
+  # and Mydia clients cannot connect a Trakt account.
+  TRAKT_CLIENT_ID: "your-trakt-client-id"
+  TRAKT_CLIENT_SECRET: "your-trakt-client-secret"
+
   # REQUIRED: Master pepper for rendezvous namespace derivation
   # Generate with: openssl rand -hex 32
   RENDEZVOUS_MASTER_PEPPER: "your-secure-random-pepper-here"
--- a/lib/mydia_web/live/integrations_live/index.ex
+++ b/lib/mydia_web/live/integrations_live/index.ex
@@ -9,6 +9,10 @@ defmodule MydiaWeb.IntegrationsLive.Index do
 
   require Logger
 
+  @trakt_not_configured_message "Trakt is unavailable because the metadata relay this server " <>
+                                  "uses has no Trakt credentials configured. If you run your own " <>
+                                  "relay, set TRAKT_CLIENT_ID and TRAKT_CLIENT_SECRET on it."
+
   @impl true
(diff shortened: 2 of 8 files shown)


arsfeld:
Superseded by #406, which removes the Trakt integration entirely.

Trakt limits free accounts to one connected Community App as of July 2026, and mydia's users typically already have another app connected, so provisioning relay credentials would not have made the integration usable. #406 deletes the relay's /trakt/* routes, so hardening their error handling is moot.

PR #402, opened 2026-08-11
fix(library): let users choose a movie's download library, stop new libraries capturing imports

arsfeld:
Adding a second movies library captured every subsequent movie import, including movies whose files already lived elsewhere. Reported by a user who added a "Cartoons" library alongside "Movies" and found pre-existing movies downloading into Cartoons.

## Root cause

`MediaImport.determine_library_path/1` re-derived the destination on every import with `Enum.find` over `Settings.list_library_paths/1`, which orders `monitored DESC, path ASC`. The effective rule was "the monitored, type-compatible library whose path sorts alphabetically first", so `/media/Cartoons` beat `/media/Movies` for everything.

Two further defects surfaced while fixing it:

- The Add Movie library dropdown was write-only on **both** paths. `build_media_item_attrs/3` never read `config.library_path_id`, and `media_items` had no column to store it in.
- "Configure Before Adding" crashed. `validate_config/2` built a schemaless changeset seeded with `%{}`, so `apply_changes/1` dropped any field submitted blank. A blank quality profile or an unticked Monitor raised `KeyError` and killed the LiveView mid-add.

## Approach

New `Mydia.Library.TargetResolver` is the single authority for "which library does this item's content go into", resolving in order:

1. `:download_override` - the download row's own `library_path_id` (existing specialized-library escape hatch, unchanged)
2. `:explicit` - the media item's new `library_path_id`
3. `:existing_files` - where this item's non-trashed files already live
4. `:type_default` - the operator-set default for the item's kind
5. `:first_compatible` - the historical alphabetical rule

Step 5 is deliberately the pre-existing behaviour, so an install that never configures a default resolves exactly as it did before.

Each candidate is re-validated before acceptance and falls through if stale, since a library can be deleted, disabled or retyped long after an item started pointing at it. The check differs by step on purpose: an explicit choice does not require `monitored`, because unmonitoring usually means a drive is offline and silently rerouting an explicit choice to a different disk is how one movie ends up in two places.

## Data model

- `media_items.library_path_id`, nullable, `on_delete: :nilify_all`. NULL means "no preference, resolve dynamically".
- `library_paths.default_for_movies` / `default_for_series`. Two flags rather than one because a `:mixed` library serves both kinds; with a single flag a flagged `:movies` library and a flagged `:mixed` library would both be "the default" and the tie would be arbitrary. Enforced by changeset validation plus a partial unique index each.
- Backfill flags the oldest monitored library per kind, reproducing most installs' behaviour before the change. Configurable via `LIBRARY_PATH_<N>_DEFAULT_FOR_MOVIES` / `_DEFAULT_FOR_SERIES` and in Admin, Library Paths.

## UI

- Discover, dashboard and trending modal: the add button becomes a split button. Plain click uses the default, the caret picks a library. The caret is hidden when fewer than two candidates exist, so single-library installs are unchanged.
- Media detail page: a Library row showing the destination and why it resolved that way ("matches existing files", "library default", "chosen"), editable. Changing it redirects future grabs only; files already on disk never move.

## Also fixed

- **Add + Search never searched.** `search_on_add` was collected on both paths and threaded into `config`, then never read. Now calls `Search.queue_auto_searches/1`, the same Oban-dedupe-safe path the detail page uses.
- **The library picker would have been a no-op for TV shows.** `add_tv_show_to_library/3` did not accept opts, so a caret pick could not reach `build_media_item_attrs/3` for series.

## Compatibility

All three migrations are additive, no `ALTER COLUMN`, so no SQLite table rebuild. Nothing is required of the operator: an install that never opens the new settings keeps working via step 5. No GraphQL schema or player changes, so there is no cross-version window to manage.

## Testing

- `TargetResolver`: 14 tests covering each resolution step, fall-through on disabled/retyped/unmonitored candidates, the explicit-but-unmonitored case, most-files-wins tie-breaking, and trashed-file exclusion.
- Regression test reproducing the reported bug directly: two movies libraries where the alphabetically-first is not the one holding the files.
- Backfill, changeset invariants, the gear-modal `KeyError` repro, the Discover picker, the detail-page row, and Add + Search.
- Full suite green on SQLite (6876 tests). Schema and migration tasks verified on PostgreSQL.

## Out of scope

`GET /api/v1/media/:id` and `POST /api/v1/media/:id/match` return 500 today: both preload `:library_path` on `MediaItem`, and `serialize_media_item/1` reads six fields that live inside `metadata` rather than on the struct. This PR adds `belongs_to :library_path`, which incidentally fixes the preload, but the serializer still raises. Broken before and after, unchanged in severity, and nothing in the repo calls those endpoints. Left for a separate change.

### DIFF
--- a/lib/mydia/config/loader.ex
+++ b/lib/mydia/config/loader.ex
@@ -512,6 +512,16 @@ defmodule Mydia.Config.Loader do
         System.get_env("#{prefix}CATEGORY_PATHS"),
         &parse_json/1
       )
+      |> put_if_present(
+        :default_for_movies,
+        System.get_env("#{prefix}DEFAULT_FOR_MOVIES"),
+        &parse_boolean/1
+      )
+      |> put_if_present(
+        :default_for_series,
+        System.get_env("#{prefix}DEFAULT_FOR_SERIES"),
+        &parse_boolean/1
+      )
     end)
     |> Enum.reject(&(&1 == %{}))
   end
--- a/lib/mydia/config/schema.ex
+++ b/lib/mydia/config/schema.ex
@@ -219,6 +219,8 @@ defmodule Mydia.Config.Schema do
       field :type, Ecto.Enum, values: [:movies, :series, :mixed, :music, :books, :adult]
       field :monitored, :boolean, default: true
       field :scan_interval, :integer
+      field :default_for_movies, :boolean, default: false
+      field :default_for_series, :boolean, default: false
     end
 
     # Env/YAML-sourced installed plugins (PLUGIN_<N>_*). DB-sourced installs
@@ -562,7 +564,9 @@ defmodule Mydia.Config.Schema do
       :path,
       :type,
       :monitored,
-      :scan_interval
+      :scan_interval,
+      :default_for_movies,
+      :default_for_series
     ])
     |> validate_required([:path, :type])
     |> validate_inclusion(:type, [:movies, :series, :mixed, :music, :books, :adult])
--- a/lib/mydia/downloads/queue.ex
+++ b/lib/mydia/downloads/queue.ex
@@ -741,20 +741,13 @@ defmodule Mydia.Downloads.Queue do
   end
 
   defp resolve_rematch_destination(media_item, episode) do
-    # Build a probe reflecting the NEW target so determine_library_path resolves
-    # the correct destination (the download row still holds the old target here).
-    probe = %Download{
-      media_item_id: media_item && media_item.id,
-      media_item: media_item,
-      episode_id: episode && episode.id,
-      episode: episode,
-      library_path: nil,
-      library_path_id: nil
-    }
+    # Resolves against the NEW target: the download row still holds the old one
+    # here, so no download is passed and step 1 is skipped by design.
+    _ = episode
 
-    case Mydia.Jobs.MediaImport.determine_library_path(probe) do
-      nil -> {:error, :no_library_path}
-      library_path -> {:ok, library_path}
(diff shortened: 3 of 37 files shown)


arsfeld:
All five addressed in 8a65ee82.

**Download override with an unpreloaded association** — fixed. `download_target/1` now falls back to `Repo.get(LibraryPath, id)` when `library_path_id` is set but the association is not loaded. Worth noting the production path was never exposed: `MediaImport.determine_library_path/1` short-circuits on `download.library_path` before reaching the resolver, so only a direct `resolve/2` call passing `download:` could hit it. It was still a real trap for future callers. Added a test that passes a `%Ecto.Association.NotLoaded{}` and asserts the override wins.

**Existing-files tie-breaking** — fixed. Added `asc: f.library_path_id` as a tertiary sort key, so an exact tie on both count and `max(inserted_at)` resolves identically on every run.

**Dynamic atom construction in `phx_values/1`** — fixed by removing the helper. `library_picker_menu` now takes explicit `tmdb_id` and `media_type` attrs and emits the `phx-value-*` bindings directly. One correction to the reasoning: the keys were never attacker-controlled, they came from our own call sites as compile-time literals, so this was not a DoS vector. It was still worth removing as a latent trap and because the codebase prohibits building atoms from runtime values.

**Blank `library_path_id` in DiscoverLive and DashboardLive** — fixed in both. `""` is truthy in Elixir, so the old `if library_path_id` guard would have forwarded an empty string into the changeset and failed the foreign key rather than falling back to normal resolution. Both now match on `is_binary(id) and id != ""`.

PR #403, opened 2026-08-11
fix(plex): repair watch state sync and replace it with a generic sync layer

arsfeld:
Repairs Plex watched sync, which shipped in `b6fca05b` and has **never run once** in production, then replaces its Plex-only engine with a provider-generic one.

## Why it never ran

Three independent, individually fatal blockers, none of which surfaced anywhere. Found by inspecting the live instance:

- **Sync was never enabled.** `connection_settings` was `NULL`, so the scheduler enqueued zero per-user jobs. All 335 "completed" Oban records are a scheduler tick finishing in ~10ms having done nothing. Individual sync jobs ever inserted: **0**.
- **The stored URL was unroutable.** A single address frozen at setup time, on a network the server no longer has an interface on. Every call burned ~27s in Req retries. `PlexOAuth.best_connection_url/1` sorted by "local HTTPS first" and never probed reachability, and `parse_server/1` discarded every other advertised connection.
- **The token was revoked, and the health check could not tell.** `test_connection/1` probed `/identity`, which requires no authentication. Verified: it returns HTTP 200 with a garbage token and with no token header at all. Connection tests and the health badge validated nothing beyond "a Plex server answers here".

Each is a general defect, not a local misconfiguration. Any self-hoster whose Plex address changes, whose token is revoked, or who never finds the toggle gets the same silent nothing.

## What changed

**Connectivity that self-heals.** Configs now store every advertised connection plus the stable `machine_identifier`, and resolve a working endpoint at call time by probing candidates concurrently. When all candidates fail, addresses are re-fetched from plex.tv and matched by `machine_identifier`, so a server that moved networks is rediscovered with no operator action.

**Auth failures that say so.** `test_connection/1` moved onto an authenticated endpoint and returns a classified `:auth` / `:unreachable` / `:unexpected` error. Auth failures surface a reconnect prompt that updates the existing config in place, preserving user links and sync state.

**Silence made impossible.** A provider-agnostic `sync_runs` table records every attempt, including skips with reasons (`sync_disabled`, `server_disabled`, `no_user_mapping`, `link_token_missing`). The 335 invisible no-ops would have been 335 visible skips. `last_watched_sync_at` now stamps only on success.

**Per-user identity.** `media_server_user_links` binds each Mydia user to an individual Plex account, seeded from Plex Home with per-profile tokens minted via user switch. A user whose token cannot be minted gets **no link at all** rather than one carrying the admin token, and unmapped users are skipped visibly instead of silently merged.

**A generic sync layer.** `Mydia.WatchSync` replaces `MediaServer.WatchedSync`, which is deleted outright rather than deprecated so any unmigrated call site fails to compile. Two tables (`remote_item_mappings`, `watch_sync_states`) and a pure three-way reconciler over local state, remote state, and the last-agreed snapshot. Jellyfin, Trakt, and plugins can implement `WatchSync.Provider` without touching the engine.

**Unwatch and resume position.** The snapshot table is what makes unwatch expressible: unwatching deletes the progress row, so local state is otherwise indistinguishable from "never watched". Positions sync both ways via `viewOffset` and `/:/progress`. `delete_progress/3` now emits `playback.unwatched`, which previously emitted nothing at all.

**Incremental sync.** The full library crawl runs on first sync and on demand, not every 30 minutes. Subsequent runs use `lastViewedAt>` filtering and container paging, with the cursor deliberately trailing the last write so a view landing mid-run is not skipped forever.

## Plugin ABI

`mydia:plugin@1.2.0`, additive only: `position-seconds` / `duration-seconds` on `playback-progress`, plus `set-watch-state`, which supersets `ensure-watched` with an unwatch path. The host publishes `host@1.2.0` alongside `1.1.0` and `1.0.0`, so existing compiled guests keep linking. No coordinated deploy order is required.

## Notable fixes found along the way

- `JsonListType` deliberately tolerates legacy cardigann cookie shapes; tightening it would break live rows, not just tests.
- Making `url` nullable for Plex discovery required a per-type addressability rule, since Jellyfin has no discovery and calls `String.trim_trailing(config.url, "/")`.
- The watched-sync toggle rendered as disabled when enabled, because the row compared `== true` against a value the form submits as the string `"true"`, which also hid the "Sync Now" button.
- The direction `<select>` set a `value` attribute browsers ignore, so it always displayed "Bidirectional".
- A new event type needs registering in three places; missing `Events.Presentation` makes it validate-fail and never broadcast.

## Testing

- Full suite green on SQLite and PostgreSQL.
- All five migrations roll back and re-apply, including the SQLite table rebuild for the nullable `url`.
- `mix compile --warnings-as-errors` passes, which is the gate proving nothing still references the deleted engine.
- Reconciler covered by an exhaustive truth table, including a property test that both sides moving off the snapshot can only converge. A "both changed in opposite directions" conflict is unreachable for a boolean and the branch was removed.
- Regression test asserting a server that answers `/identity` but 401s on real endpoints reports as an auth failure.
- Regression test asserting a disabled sync records a `skipped` run, the exact defect that hid for eight months.


arsfeld:
All five addressed in 8dc1ee8f. Notes on each, including where the reasoning differed from the diagnosis:

**1. `apply_link_token/2` cannot verify link ownership.** Correct, and it matters more than the crafted-job framing suggests: this PR exists to stop separate users' watch state being merged, and resolving a token by link id alone reintroduces exactly that from the other direction. It now takes `user_id` and pattern matches `%MediaServerUserLink{user_id: ^user_id}`, returning `:link_user_mismatch` otherwise, which is recorded as a skip like any other. Regression test added.

**2. `{:ok, run} = start_run(...)` MatchError.** Correct. Bookkeeping exists to explain what happened, so losing it should degrade observability rather than fail the sync. `start_run/3` now returns the run or `nil` on failure with a logged warning, and `finish_run/4` has a `nil` clause.

**3. `Task.async_stream` ordering.** Right conclusion, slightly different mechanism. `ordered: true` does not serialize the probes; they still run concurrently up to `max_concurrency`. The actual cost was that `Enum.flat_map` materialized every result before `Enum.find` picked a winner, so resolution always waited for the slowest dead candidate even when the first answered instantly. Now `ordered: false` with a lazy `reduce_while` that halts on the first success, while still collecting observed failures so error classification keeps working.

**4. `acc ++ items` is O(n^2).** Correct, and it lands on the initial crawl, which is the worst place for it. Now prepends pages and reverses once at the end.

**5. Direction is never enforced.** Correct, and the most user-visible of the five: import-only and export-only were read from settings, recorded on the run, and then ignored. `direction` is now threaded into `WatchSync.Engine`, which gates the propagating decisions. It is enforced at the engine rather than the reconciler so the reconciler stays pure and keeps reporting what *should* happen; a direction-suppressed item still updates its snapshot so the last-agreed state stays truthful and the next run does not see phantom drift. Two tests cover it.

PR #409, opened 2026-08-11
feat(plugins): sync the library to Simkl plan-to-watch and pull back favorites

arsfeld:
Extends the bundled Simkl plugin past watched state to the list surfaces the original brainstorm deferred, mapping onto Mydia concepts that already exist. No new schema.

## What it does

- **Push:** owned-but-unwatched library items go to Simkl as `plantowatch`.
- **Pull:** Simkl `plantowatch` items Mydia knows about but does not own become Favorites for that user.

## Why it is shaped this way

Measuring the deferred surfaces against the real Simkl API collapsed two of the three:

- **Simkl has no owned-media concept.** There is no `/sync/collection` and no equivalent. The only honest translation of Trakt's collection push is "your library shows up as plan-to-watch", which makes collection sync and watchlist sync the same feature here.
- **`watching` and `completed` need no work.** Simkl derives both from `/sync/history`, which this plugin already drives. Of the five Simkl statuses, `plantowatch` is the only one not already covered that can be derived from local state.

## The echo guard is a set difference

Push and pull would otherwise form a loop: push claims every unwatched library item, then the next pull sees exactly those on `plantowatch` and favorites them, turning a hand-curated collection into a copy of the library.

Rather than remembering what it pushed, the plugin computes two sets each run and takes the difference in both directions:

- **D** = owned locally with nothing watched by this user
- **P** = Simkl's `plantowatch` list (`extended=simkl_ids_only`, one cheap fetch)

Push `D \ P`, favorite `P \ D`. The two are complementary by construction, so the plugin cannot favorite its own output, and there is no KV state to lose. Convergence falls out of the same property: once synced, both diffs are empty and the leg writes nothing.

A stored guard was the obvious alternative and is worse on both counts. `Mydia.Plugins.KV` caps a value at 64 KB, roughly 3000 keys, so a large library silently overflows it; and losing that key to a reinstall or a reconnect floods Favorites with hundreds of rows the user then deletes by hand.

Consequence worth knowing: an item you own unwatched *and* added to `plantowatch` on Simkl yourself will not become a local favorite. The guard cannot tell that apart from its own output. It is a false negative, never a false positive.

## Contract

Bumps to `mydia:plugin@1.2.0`, additive only per the evolution rules in `plugin.wit`:

- New record `library-item` and a new `list-item` variant case, behind a new `data:read` namespace `library_item`. This exists because `media-item` carries metadata but cannot answer "do we have the file", which is what distinguishes a catalogued item from an owned one. It is a separate record rather than a new field, since adding a field to an existing record would break 1.1 guests.
- New `ensure-favorite` host function behind `surfaces:write` scoped to `collections:favorite`, modeled on `ensure-watched`: host-side external-id matching, an active connection to the target user required (R21), idempotent. Deliberately additive with no remove counterpart, so a remote list can never delete local curation.

Both bundled 1.1-era guests still compile and instantiate against the 1.2 host, which is the proof the change stayed additive.

## Not included, and why

- **Ratings.** Mydia has no user rating, and Simkl cannot rate seasons or episodes even if it did.
- **`hold` and `dropped`.** Nothing local to derive them from.
- **Removal in either direction.** `POST /sync/history/remove` is the only removal path and it wipes the item's rating as a side effect. Deleting a file is not a statement about intent, so a local deletion never reaches Simkl.

## Operator impact

The manifest widens `data:read` and `surfaces:write`, so an already-approved install keeps running on its narrower grant and the new calls return `Denied` until the operator re-approves. That is the platform's existing behavior and the admin UI already surfaces it.

Two new settings, `sync_plantowatch` and `sync_favorites`, both default on: connecting a Simkl account is itself the consent, and the plugin is opt-in per user. They are `enum` rather than boolean because `Manifest.@setting_types` has no boolean type.

## Testing

- 29 Rust guest unit tests covering the mini-list parser, the `add-to-list` body builder, the intent set, and that the two diffs are complementary and converge
- Host tests for the `library_item` namespace (capability denial, pagination, `owned` correctness across movie/show with and without files) and for `ensure_favorite` (grant denial, connection scoping, idempotency, not-found)
- An end-to-end Bypass test proving push then pull does not favorite its own output
- Ownership query verified on both SQLite and PostgreSQL
- Full precommit green: 6902 tests, 0 failures, Credo strict clean


arsfeld:
fix(simkl): address Copilot review on the list leg

All three findings were real.

  * data-list errors were swallowed. Both list_library and list_progress ended
    a page loop on any error, so a Denied (an operator has not re-approved the
    widened grant yet) or a transient failure produced a partial library rather
    than a failure. That is not a smaller answer, it is a wrong one: an empty D
    makes P \ D the user's entire Simkl plan-to-watch list, so the leg would
    favorite all of it. Exactly the flood the set-difference guard exists to
    prevent. Both now propagate.

  * The full progress scan ran once per connection, on top of the history
    leg's incremental scan. data-list is consent-scoped host-side and
    local_intent_set already filters by user, so one scan per tick is hoisted
    into on_schedule and shared across connections.

  * apply_favorite checked is_favorite? and then inserted, which races the
    collection_items unique index against a user starring the same item. The
    constraint violation is the same outcome the check guards against, so it
    now reports already-favorited instead of an internal error.

The imports_for test asserted ensure-favorite under the 1.2.0 namespace key,
which the merge made wrong: 1.2.0 is deliberately narrowed to exclude it. Fixed,
and a test added that locks the narrowing itself, since handing an older guest a
function its own contract never declared fails wasmtime's import check.

PR #415, opened 2026-08-12
fix(requests): one-click guest requests and posters on approved items

arsfeld:
Fixes two defects in the guest request path.

- Discovery's Request button navigated guests to an empty search form. It built `/request/movie?tmdb_id=<id>`, but `RequestMediaLive` only reads `?q=`, so the id was silently dropped. It now submits the request in one click, mirroring the admin Add to Library button
- Approving a request created a media item with no `metadata`, so it rendered an empty poster placeholder forever. Approval now fetches provider metadata before opening the transaction, through the same `Mydia.Media.Add` path the admin add flow uses
- Extracted `Mydia.Media.Add` from `MediaAddHelpers` so there is one path from a provider ID to a library row, shared by the add flow and request approval
- Added `Mydia.Jobs.MetadataBackfill` to repair items already stored without metadata, daily at 5:15 AM

If the metadata relay is unreachable at approval time the request stays pending and the admin sees an error, rather than a metadata-less item being created.

Verified on SQLite (7005 tests, 0 failures) and PostgreSQL (53 affected tests, 0 failures), `credo --strict` clean.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -300,6 +300,8 @@ config :mydia, Oban,
        {"*/5 * * * *", Mydia.Jobs.SegmentDetectionScheduler},
        # Refresh metadata for all monitored items weekly on Sunday at 5 AM
        {"0 5 * * 0", Mydia.Jobs.MetadataRefresh, args: %{"refresh_all" => true}},
+       # Repair media items stored with no metadata (blank posters) daily at 5:15 AM
+       {"15 5 * * *", Mydia.Jobs.MetadataBackfill},
        # Permanently delete trashed media files past retention period daily at 5 AM
        {"0 5 * * *", Mydia.Jobs.TrashCleanup},
        # Sync runs accrue per server per user per tick, so they need pruning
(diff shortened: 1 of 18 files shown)


arsfeld:
Thanks, all three handled in 6e926b8e.

**`String.to_existing_atom/1` on `phx-value-media_type`** (both LiveViews): fixed. One clarification on severity: the raise kills only the sender's own LiveView process, so it is a self-inflicted crash by an authenticated user rather than a DoS against other users. It is still a real crash that lands in our crash reporting, so it is worth closing. Discover and Dashboard now match the two known types explicitly and flash an error otherwise.

I applied the same treatment to the adjacent `add_to_library` and `show_details` handlers, which had the identical pattern and were not part of this diff. Leaving one hardened and its twin two lines below unhardened seemed worse than the slightly wider diff. Dashboard's `show_details` also had a `case media_type do` with no catch-all, which would have raised a `CaseClauseError` on the same input; that is gone too.

Note the strict parser is named `parse_event_media_type/1` rather than reusing Discover's existing `parse_media_type/1`. The latter handles the `?type=` URL param and deliberately falls back to `:movie` for unknown input, which is the right behaviour for a user-typed URL. Client event payloads want the strict version.

**`fetch_provider` naming**: agreed and renamed to `primary_provider`, with the comment corrected. You were right that the fetch is unconditionally TMDB and the value selects the primary content provider downstream. That name came across verbatim in the extraction from `MediaAddHelpers`.

Verified: 7005 tests 0 failures, `credo --strict` clean, `compile --warnings-as-errors` clean.

PR #417, opened 2026-08-12
fix(player): let the Flatpak player persist a pairing

arsfeld:
A pairing made in the Linux Flatpak player did not survive a restart.

## Cause

Every persisted value in the player goes through `flutter_secure_storage`. On Linux that reaches the host keyring via libsecret, which talks to the D-Bus Secret Service at `org.freedesktop.secrets`. The Flatpak manifest granted no session bus names beyond `org.freedesktop.ScreenSaver`, and a sandbox denies every name that is not listed, so libsecret could not reach the service.

`_NativeAuthStorage._withFallback` then caught the failure and redirected the value to a process-local map behind a single `debugPrint`. The write looked successful to every caller, worked for the session, and was gone on the next launch.

Confirmed on a real install rather than inferred from the manifest:

- `flatpak info --show-permissions dev.mydia.player` listed only `org.freedesktop.ScreenSaver=talk`
- the host had a working Secret Service the whole time (gnome-keyring owning `org.freedesktop.secrets`)
- a D-Bus ping inside the sandbox returned `ServiceUnknown`
- the same ping with `--talk-name=org.freedesktop.secrets` returned `()`

## Blast radius

Pairing was the visible symptom, but everything sharing that storage was affected and reset on each launch: cert fingerprints, media tokens, player settings (default quality, auto-skip, library sort), the update-check timestamp, and `device_id`. Because `device_id` was regenerated every launch, each pairing attempt registered a **new** device server-side.

## Changes

- Add `--talk-name=org.freedesktop.secrets` to the manifest `finish-args`
- Add `player/flatpak/test/finish-args-test.sh`, asserting the manifest declares every runtime permission the app depends on, each with the feature that breaks without it; wired into the seconds-long `flatpak-validate` job rather than the 90-minute build
- Assert `org.freedesktop.secrets=talk` on the exported artifact in `smoke-test.sh`, which catches a permission lost between manifest and export
- Track a `degraded` flag on `AuthStorage` when a write fails to reach durable storage, keeping the in-session fallback so the app still works
- Surface that through `AuthService.storageDegraded` and `LoginState.credentialsNotPersisted`
- Warn the user with a dialog they must acknowledge before entering the app, via a single `_completePairing()` helper that also collapses three copy-pasted navigation tails in `login_screen.dart`

The silent degradation is the reason this shipped unnoticed. The same shape already cost macOS once, documented in `secure_storage_options.dart`. A lost pairing now says so instead of looking identical to a saved one.

## Testing

- Full player suite green (2098 tests, `--concurrency=1`)
- `finish-args-test.sh` verified in both directions: it fails on the pre-fix manifest with the exact missing-permission message, and passes after
- The smoke-test assertion verified to fail against a pre-fix install, so it is not passing vacuously
- New unit tests cover degraded-on-write, not-degraded-on-failed-delete (guarding the macOS -34018 tolerance), stickiness, and the in-session fallback

Not covered here: pairing against a live server and confirming the restart survives end to end needs a human and a running instance.

### DIFF
--- a/.github/workflows/ci-player.yml
+++ b/.github/workflows/ci-player.yml
@@ -327,6 +327,9 @@ jobs:
       - name: Validate AppStream metainfo
         run: appstreamcli validate --explain player/flatpak/dev.mydia.player.metainfo.xml
 
+      - name: Validate finish-args
+        run: player/flatpak/test/finish-args-test.sh
+
   flatpak-build:
     name: Flatpak / Build
     runs-on: ubuntu-latest
--- a/player/flatpak/dev.mydia.player.yml
+++ b/player/flatpak/dev.mydia.player.yml
@@ -43,6 +43,11 @@ finish-args:
   - --device=dri
   # Stop the screen blanking during playback.
   - --talk-name=org.freedesktop.ScreenSaver
+  # flutter_secure_storage reaches the host Secret Service through libsecret,
+  # which talks D-Bus. Without this the sandbox reports the name as unknown,
+  # every credential write fails, and a pairing silently does not survive a
+  # restart. player/flatpak/test/finish-args-test.sh guards this line.
+  - --talk-name=org.freedesktop.secrets
   # The Flutter bundle and mpv both live under /app/lib.
   - --env=LD_LIBRARY_PATH=/app/lib
   # No --filesystem: the player streams from a server and reads no local media.
--- a/player/flatpak/smoke-test.sh
+++ b/player/flatpak/smoke-test.sh
@@ -39,4 +39,23 @@ if grep -qiE 'error while loading shared libraries|cannot open shared object' "$
   exit 1
 fi
 
+# The Secret Service permission does not crash anything when missing. It
+# silently costs the user their pairing on the next launch, which no runtime
+# smoke test can observe in ten seconds. Assert on the exported metadata
+# instead, which is what actually governs the installed app.
+#
+# This reads metadata rather than making a live D-Bus call on purpose: a real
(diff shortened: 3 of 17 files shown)


arsfeld:
All five review comments addressed in 245eb724.

**The stale-read one was a real bug, thanks.** `read()` consulted the backend first and only fell back to memory when that returned nothing, so a write that had failed and landed in memory was shadowed by whatever older value the keyring still held. Concretely: a refreshed access token that failed to persist would leave every later read handing back the expired one for the rest of the session, which defeats the point of the fallback. The original test missed it because the fake backend held no prior value for the key.

Memory is now a per-session overlay: mirrored on every write rather than only on failure, cleared on every delete and `deleteAll`, and preferred whenever it holds the key. Three regression tests added covering a failed write over an existing backend value, a successful write following an earlier fallback, and a delete clearing the overlay.

The rest:

- `storageDegraded` doc reworded. The flag latches on the first failure, and writes either side of it may still have persisted, so it now says durability can no longer be guaranteed rather than that everything is lost.
- Dialog copy changed from "will be lost" to "may be lost", for the same reason: the flag is sticky and process-wide, so the app cannot strictly know this particular connection is gone.
- Dialog explanation now names both causes, a missing keyring service and a sandbox without permission to reach one. The second is exactly the bug this PR fixes, so leading with the first was misleading.
- Stale "IPFS DHT bootstrap" comment corrected. Pre-existing and wrong since the move to iroh, per CLAUDE.md.

Full player suite green at 2101 tests.

PR #426, opened 2026-08-12
fix(downloads): give up on a stalled download instead of mislabelling it as a failed import

arsfeld:
## The problem

A stalled download surfaced this on the Downloads screen:

> Import failed: stalled after 180m without progress — escalated to failure

Four things wrong with it:

1. **It is a download problem labelled as an import problem.** The stall detector had no download-side terminal field, so it reused `import_failed_at` / `import_last_error`, and the row template prefixes any non-nil `import_last_error` with `Import failed:`.
2. **The rule was documented nowhere.** The only user-facing mention was the "Stalled timeout (minutes)" field, whose help text described one of the two thresholds.
3. **"Escalated to failure" was not true.** Escalation wrote two timestamps and stopped. The torrent kept running in the client and the row stayed in the queue. Its only real effect was freeing the episode for re-search.
4. **The number was wrong.** Soft-stall fires at 60m, escalation 180m *after that*, so a download reaching that message had made no progress for 240 minutes.

## What changed

Stall escalation stops being a bespoke third terminal path and calls `Queue.reject_release/2`, the same primitive the bad-content path already uses. Giving up now means: blocklist the release for **1 day** (short, because a dead swarm may recover overnight, unlike a corrupt release), remove the torrent from the client, delete the row, and queue a replacement search immediately. Guarded by the existing per-item `auto_reject` cap, so a media item whose detector is misfiring is left alone rather than churned.

The explanation moves earlier, to where it can still change the outcome. The soft-stall row now names the deadline and offers both ways out:

> **Stalled** · No progress for 1h 12m. If nothing changes in the next 2h 48m, Mydia will remove it and search for a different release.
> `[Remove and find another]` `[Keep waiting]`

"Keep waiting" is a snooze, not an opt-out: it buys one full fresh window.

The client config field now derives and shows both thresholds instead of only the first, which makes the debrid default legible for the first time (flagged after 24h, given up on after 4 days).

`docs/using/reference/download-clients.md` gains a "Stalled Downloads" section covering both stages, the auto-reject cap, and the observation-gap reset. That last one is the least discoverable behaviour in the system: stall time only accrues while Mydia is actually watching, so a restart, a client outage, or a paused torrent resets the clock rather than accruing toward a give-up.

## Along the way

`reject_release/2` gains `:ttl_days` and `:event` (both defaulting to today's behaviour), and stops refusing outright when a download has no `(indexer, guid)` key. Previously it returned `{:error, :no_guid}` and changed nothing, leaving the torrent running. There is nothing to blocklist without a key, but there is still a dead download to clear. This also unblocks the reject button in the match-files modal, which was disabled for exactly those downloads.

The escalation message now reports the true total and drops the false claim: `"no progress for 4h"`.

Removed as unreachable: `StallDetector.stalled?/1`, the LiveView's `stalled?/1`, the `stall_failed` badge class, and the "Stall failed" badge branch.

## Verification

- Full suite: 7152 tests, 1 failure
- `credo --strict`: 13608 mods/funs, no issues
- `format --check-formatted` and `compile`: clean

The single failure is `Mydia.Indexers.Adapter.ProwlarrTest` "parses quality from title", which is **pre-existing and unrelated**. This PR touches no indexer file; the test passes in isolation; and it is a placeholder asserting `function_exported?(Prowlarr, :search, 3)` without ensuring the module is loaded, which returns false for an unloaded module and so races in a parallel run. Worth a separate one-line `Code.ensure_loaded?/1` fix.

### DIFF
--- a/docs/using/reference/download-clients.md
+++ b/docs/using/reference/download-clients.md
@@ -45,3 +45,37 @@ These fields appear when adding or editing a client in the Admin UI. Not every f
 | Category | Default category | `mydia` |
 | Priority | Client priority | `1` |
 | Download Directory | Output directory | `/downloads` |
+| Stalled Timeout | Minutes without progress before a download is flagged as stalled | `60` |
+
+## Stalled Downloads
+
+Mydia watches byte progress on every active download and acts in two stages.
+
+**Stage 1: flagged as stalled.** A download that makes no byte progress for the client's **Stalled Timeout** (default 60 minutes) is flagged with a yellow "Stalled" badge on the Downloads screen. This is a warning, not a verdict. The download keeps its place in the queue, and if bytes start moving again the flag clears itself.
+
+**Stage 2: given up on.** If the download stays stalled for three times the timeout after that (180 minutes on the default, so **4 hours without progress in total**), Mydia gives up on the release. It:
+
+- removes the torrent from your download client
+- blocklists that release for **1 day**, so the replacement search does not immediately grab the same dead copy back
+- deletes the download from the queue
+- queues a fresh search straight away
+
+The blocklist is deliberately short. A torrent with no seeds today may have seeds tomorrow, so a stall earns a cooldown rather than the near-permanent block a corrupt release gets. You will find the give-up recorded in the activity feed and on the media item's history.
+
+While a download is in stage 1, the row tells you exactly when stage 2 will fire and offers two buttons: **Remove and find another** does it immediately, and **Keep waiting** resets the clock and buys another full window.
+
+### Restarts, outages, and paused torrents do not count
+
+Stall time only accrues while Mydia is actually watching a download run. If it goes unobserved for more than 6 minutes, for any reason, the clock resets instead of accruing.
+
+That covers a Mydia restart, a download client that goes unreachable, and a torrent you paused or that sat queued behind others. It is why a torrent can look frozen at the same byte count for a day and never be touched: Mydia only counts the stretches where it watched the download fail to progress, not wall-clock time.
+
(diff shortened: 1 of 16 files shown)


arsfeld:
All three review comments were valid and are fixed in 2b5f48ab.

**1. `download.failed` emitted before the reject** — correct, and the worse half is the repeat. On a persistent reject failure the row survives with `stalled_since` and `import_failed_at` untouched, so the next poll re-enters the same branch and emits a second terminal event for a download still sitting in the queue. Moved the emit into the `{:ok, :rejected}` branch. The in-memory struct outlives the deleted row, so nothing is lost by waiting. That branch now returns `0`, since nothing changed means nothing newly stalled.

**2. Warning noise on missing blacklist keys** — agreed. Now split: `:no_indexer` / `:no_guid` log at debug, because having no release key is an ordinary shape for externally-adopted torrents and manual grabs rather than a fault. A blacklist write that genuinely fails still warns, since that release will not be filtered out of the next search.

**3. Suppression event repeating** — real, though slightly narrower than described. It does not repeat forever: `suppress_stall_reject/2` leaves `last_observed_at` stale, so after ~6 minutes the observation-gap reset clears `stalled_since` as a side effect and stops it. But that means bursts of roughly 24 duplicate events while the 15s fast-followup chain is running, and relying on an unrelated safety net to stop an event loop is fragile. It now resets the stall clock explicitly, the same reset the operator's "Keep waiting" button performs, giving one event per full grace+escalation cycle.

Added `AE6d` to lock point 3 in: two polls across a suppressed give-up must produce exactly one `download.auto_reject_suppressed` event.

PR #453, opened 2026-08-15
feat: TMDB recommendations rail on the detail page and Discover modal

arsfeld:
Adds a "More like this" rail to the media detail page and the Discover detail modal, sourced from TMDB's behavioural recommendations ("people who liked X liked Y").

## What this does

- Owned titles appear badged and link to their library page; unowned titles get the existing add or request action.
- The rail is absent entirely when TMDB returns nothing, which is common for obscure titles. No genre-based filler.
- In the Discover modal the rail is chainable: open a title, walk its neighbours.

## Why it needs no relay change

TMDB exposes recommendations as a sub-object of the details endpoint via `append_to_response`, which the relay already proxies. Reading it there rather than adding a dedicated relay route means this works on every existing install immediately, with no `metadata-relay-v*` release and no cross-version window.

The sub-object uses the same `{page, results, total_pages}` envelope as search and trending, so the existing `parse_search_results/2` consumes it unchanged.

## Notes

- Mydia already requested `recommendations` in `@shared_append_to_response` and parsed it down to bare ids in `recommended_tmdb_ids`, discarding titles and posters. Those ids fed only the GraphQL `similar` field, which intersects with the library and so returns only titles you already own. That field is unchanged here.
- No GraphQL or player changes, deliberately, to stay clear of the Rust SDL parity gate.
- 24h ETS cache keyed by title, media type and language.

## Also fixed

Ten docstrings in `lib/mydia/metadata.ex` pointed at `metadata-relay.dorninger.co`, which no longer runs the relay and today answers with an unrelated service that silently ignores `append_to_response`. They now point at `relay.mydia.dev`.

## Testing

Full suite: 7828 tests, 4 failures. Those 4 (TVShowSearch, MovieSearch, MetadataEnricher) pass in isolation (107 tests, 0 failures) and are cross-test pollution unrelated to this branch, which only adds functions to `relay.ex` and `metadata.ex` without modifying existing ones.

New coverage: provider (6), cache wrapper (4), context (8), rail component (5), detail page (3), Discover lookup (4).

## Summary by CodeRabbit

- **New Features**
  - Added recommendation rails to Discover details and media pages.
  - Browse, select, navigate to, add, or request recommended movies and TV shows directly.
  - Added language-aware recommendation results with 24-hour caching.
  - Guest users can request recommendations where supported.

- **Bug Fixes**
  - Gracefully handles unavailable, unsupported, empty, or invalid recommendations.
  - Preserves library status, request state, and existing media links for recommended items.

### DIFF
--- a/lib/mydia/media/recommendations.ex
+++ b/lib/mydia/media/recommendations.ex
@@ -54,9 +54,31 @@ defmodule Mydia.Media.Recommendations do
   def for_tmdb_id(nil, _media_type, _config), do: :none
 
   def for_tmdb_id(tmdb_id, media_type, config) when media_type in [:movie, :tv_show] do
+    case normalize_tmdb_id(tmdb_id) do
+      {:ok, tmdb_id} -> fetch(tmdb_id, media_type, config)
+      :error -> :none
+    end
+  end
+
+  def for_tmdb_id(_tmdb_id, _media_type, _config), do: :none
+
+  # `""` and `"abc"` would otherwise be interpolated straight into a relay path
+  # and spend a request to learn what the shape already tells us.
+  defp normalize_tmdb_id(id) when is_integer(id) and id > 0, do: {:ok, to_string(id)}
+
+  defp normalize_tmdb_id(id) when is_binary(id) do
+    case Integer.parse(id) do
+      {parsed, ""} when parsed > 0 -> {:ok, to_string(parsed)}
+      _ -> :error
+    end
+  end
+
+  defp normalize_tmdb_id(_), do: :error
+
+  defp fetch(tmdb_id, media_type, config) do
     config = config || Metadata.default_relay_config()
 
-    case Metadata.fetch_recommendations_cached(config, to_string(tmdb_id), media_type: media_type) do
+    case Metadata.fetch_recommendations_cached(config, tmdb_id, media_type: media_type) do
       {:ok, []} ->
         :none
 
@@ -71,6 +93,4 @@ defmodule Mydia.Media.Recommendations do
         :none
     end
   end
-
-  def for_tmdb_id(_tmdb_id, _media_type, _config), do: :none
 end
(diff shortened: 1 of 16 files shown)


arsfeld:
Thanks — all three findings verified against the code and addressed in 10a74766, along with four more from an independent review pass.

**Confirmed and fixed**

1. **Detail page emitted unhandled events (`show.html.heex`).** Correct, and worse than described: the rail reused `trending_card`, whose buttons emit `add_to_library`/`request_media`, while `MediaLive.Show` defines neither. The first click on any unowned recommendation raised `FunctionClauseError` and killed the LiveView. `add_recommendation` was unreachable dead code with no emitter. Fixed by making the action events attrs (`add_event`/`request_event`) that default to Discover's current values, so Discover is unchanged, and pointing the detail page at its own handlers. Guests also had no request path there at all; they do now.

2. **Discover rail state not refreshed (`index.ex`).** Correct on both counts. The guest case was the sharper one: `handle_info({:request_media, ...})` resolved against `items` only, so a Request click from inside the modal silently did nothing. That now resolves against the rail too, and both the add and request success paths re-enrich `selected_recommendations`.

3. **Invalid TMDB ids (`recommendations.ex`).** Correct. `""`, `"abc"` and non-positive ids were interpolated into a relay path and cost a request to learn what the shape already told us. Now rejected up front, with coverage that uses an unstubbed Bypass so the test fails if any request escapes.

**Also fixed, found independently**

- Rail items were never enriched with request status, so `requested?/1` read `nil` on every card and a guest could submit a duplicate request.
- Discover fetched recommendations synchronously inside `handle_info`, blocking the LiveView on a relay call with the modal already open; `close_details` and the action buttons queued behind it. Now `start_async`, matching the detail page.
- `decorate/2` filtered malformed ids out of the lookup but then passed the unfiltered list to `enrich_with_library_status/2`, which calls the same raising parser, so the guard did not actually hold.
- The detail rail hardcoded `libraries={[]}` (no picker on multi-library installs) and ignored `can_create_media`.

Added 7 regression tests covering the event contract and the id validation; the component test that previously waved the crash through now asserts on it.

PR #454, opened 2026-08-15
fix(metadata): let the scheduled pass refresh episodes

arsfeld:
## What

The weekly metadata pass called `Refresh.run/2` with `fetch_episodes: false`. It is the only recurring job that re-reads provider data, so episode rows were written once at import and never again. Anything a provider publishes *after* an episode airs (the screencap, overview and runtime, which TVDB fills in around the air date) could never land.

## How it showed up

On a production library, Silo S3E6 and S3E7 rendered with no thumbnail. The DB rows had `still_path`, `overview` and `runtime` all null, while a live TVDB read returned all three:

```
DB   E6 still=nil                          overview=nil
DB   E7 still=nil                          overview=nil
TVDB E6 still=.../11751884/screencap/...   overview=present
TVDB E7 still=.../11751885/screencap/...   overview=present
```

The show row had `updated_at = 2026-08-09` (the weekly pass ran, on schedule) while every one of its episode rows was still at `2026-07-28`, the import date and two weeks before those episodes aired. Not a rendering bug: the GraphQL resolver reads `still_path` correctly, and the value was simply never refreshed.

Not a regression either. `fetch_episodes: false` predates the typed-`Args` refactor, so the bulk pass has never refreshed episodes.

## The throttle was dead three ways

The first review of this PR challenged a comment claiming the new cost was bounded by `should_skip_season_refresh?/1`. That challenge was correct, and chasing it down found the throttle could never fire at all:

1. It is only reachable from `refresh_episodes_for_tv_show/2`, which the bulk pass never called.
2. It keys off `seasons_refreshed_at`, which `MediaItem.changeset/2` did not cast. Both writers (`refresh_episodes_for_tv_show/2` and `ProviderSwitch`) go through `Media.update_media_item/3`, so every write was silently dropped and the column stayed NULL. `should_skip_season_refresh?/1` returns `false` on a nil timestamp. This is why all 54 shows on the affected instance had `seasons_refreshed_at` NULL.
3. Its ended-show branch matched `%{"status" => status}` against `metadata`, which loads as an atom-keyed `%MediaMetadata{}`. `is_completed` was always `false`, so `completed_show_refresh_threshold_hours` was never read and ended shows were throttled as if still airing.

All three are fixed, so the throttle now actually runs.

## Honest cost

The throttle does **not** skip an airing show on this pass: the interval is 168h and the ongoing threshold 24h. That is the intent, since a weekly episode re-read is the point of the change. What it does buy is that another trigger within 24h (manual refresh, library scan, import) no longer redoes the work, and that ended shows fall under the longer completed-show threshold.

The residual cost is real. `season_monitoring` defaults to `"all"`, so every season of every show is re-read weekly, and each episode costs a translation fetch against the shared relay. Narrowing that to seasons that can actually have changed is the obvious follow-up, along with the duplicate series fetch below.

## Second fix: dropped relay config

`Refresh.run/2`'s episode leg called `refresh_episodes_for_tv_show/1`, which falls back to `default_relay_config/0`. An injected config was therefore silently dropped and the episode fetch escaped to the global relay. This is the same defect `refresh_episodes_config_test.exs` already pins at the other call site. Fixing it is also what makes the bug above testable without touching the real relay.

It changed one existing assertion in `media_test.exs`. The endpoint is now hit twice, because the episode leg re-fetches the show to read its season list; it only ever looked like one request because that leg escaped to the global default, meaning the routing this test exists to pin was never actually covered for the episode fetch. Per CodeRabbit, the call count is asserted explicitly (`== 2`) rather than dropped, so a third fetch cannot slip in unnoticed.

## Tests

`test/mydia/jobs/metadata_refresh_episodes_test.exs`:

- Reproduces the production shape: an episode row imported before it aired (`still_path: nil`), a provider that now has the screencap, and the pass expected to pick it up. Verified failing before the fix with the exact production symptom (`got: nil`).
- `seasons_refreshed_at` round-trips through `update_media_item/3`. Verified failing before the cast fix with `left: nil`.
- An ended show inside its threshold is not refetched. Verified failing before the `is_completed` fix.

The first version of the new stub used TMDB-shaped keys (`first_air_date`) against a TVDB endpoint, so the refresh wrote `year: nil` over the fixture while the assertions still passed. It now uses the real TVDB shape (`firstAired`, nested `status`) and asserts the year survives.

## Follow-ups, not in scope here

- The episode leg re-fetches the show payload `Refresh.run/2` just fetched, purely to read `metadata.seasons`. Pre-existing for every `fetch_episodes: true` caller; now on the weekly path too.
- `season_monitoring: "all"` re-reads every season weekly when generally only the current one can change.

## Verification

- `mix compile --warnings-as-errors` clean
- `mix format --check-formatted` clean
- `mix credo --strict` gave 14549 mods/funs, no issues
- Full suite: 7801 tests, 0 failures
- CI was green on the first commit (Test, PostgreSQL, E2E Browser, Rust, Flutter Pin, Apple Deployment Targets)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved scheduled TV episode metadata refreshes so screencaps, overviews, runtime details, and show years update correctly.
  - Ensured episode refreshes use configured relay settings.
  - Preserved NFO writing during standard media refreshes.
  - Corrected refresh throttling for completed or canceled shows and season timestamp handling, including partial or unsuccessful refreshes.

- **Tests**
  - Added regression coverage for episode metadata updates, refresh throttling, and timestamp persistence.
  - Expanded TV episode refresh routing coverage.

### DIFF
--- a/lib/mydia/jobs/metadata_refresh.ex
+++ b/lib/mydia/jobs/metadata_refresh.ex
@@ -158,8 +158,22 @@ defmodule Mydia.Jobs.MetadataRefresh do
       {:error, :not_found}
   end
 
-  defp refresh_one(media_item) do
-    Refresh.run(media_item, recover_by_title: true, fetch_episodes: false)
+  @doc false
+  # The scheduled pass is the only recurring job that re-reads provider data, so
+  # `fetch_episodes: false` here meant episode rows were written once at import
+  # and never again. Anything a provider publishes after an episode airs — the
+  # screencap, overview and runtime, which TVDB fills in around the air date —
+  # could therefore never land, leaving recently aired episodes permanently
+  # thumbnail-less while the show row refreshed weekly.
+  #
+  # Cost is bounded by should_skip_season_refresh?/1, whose thresholds (24h
+  # ongoing, 168h ended) existed for this call and were unreachable while the
+  # bulk pass never set seasons_refreshed_at.
+  #
+  # Public for testability: proving the pass requests episodes without hitting
+  # the real relay requires injecting a config.
+  def refresh_one(media_item, opts \\ []) do
+    Refresh.run(media_item, Keyword.merge([recover_by_title: true, fetch_episodes: true], opts))
   end
 
   # A single bad item must not abort the pass. The ErrorTracker report is
--- a/lib/mydia/jobs/metadata_refresh.ex
+++ b/lib/mydia/jobs/metadata_refresh.ex
@@ -166,9 +166,22 @@ defmodule Mydia.Jobs.MetadataRefresh do
   # could therefore never land, leaving recently aired episodes permanently
   # thumbnail-less while the show row refreshed weekly.
   #
-  # Cost is bounded by should_skip_season_refresh?/1, whose thresholds (24h
-  # ongoing, 168h ended) existed for this call and were unreachable while the
-  # bulk pass never set seasons_refreshed_at.
+  # On cost, precisely. should_skip_season_refresh?/1 was dead three ways: it is
+  # only reached from refresh_episodes_for_tv_show/2 (which this pass never
+  # called), it keys off seasons_refreshed_at (which MediaItem.changeset/2 did
+  # not cast, so it was never persisted and stayed nil, and a nil timestamp
+  # returns false), and its ended-show branch matched a string key against an
+  # atom-keyed struct. All three are fixed here, so the throttle now actually
+  # runs. It still does not skip an airing show on this pass — the interval is
+  # 168h and the ongoing threshold 24h — which is the intent: a weekly episode
+  # re-read is the whole point. What it does buy is that another trigger within
+  # 24h (manual refresh, library scan, import) no longer redoes the work, and
+  # that ended shows fall under the longer completed-show threshold.
+  #
+  # The residual cost is real and worth knowing: season_monitoring defaults to
+  # "all", so every season of every show is re-read weekly, and each episode
+  # costs a translation fetch against the shared relay. Narrowing that to
+  # seasons that can actually have changed is the obvious follow-up.
   #
   # Public for testability: proving the pass requests episodes without hitting
   # the real relay requires injecting a config.
--- a/lib/mydia/media.ex
+++ b/lib/mydia/media.ex
@@ -2380,10 +2380,16 @@ defmodule Mydia.Media do
   defp should_skip_season_refresh?(%MediaItem{} = media_item) do
     config = Mydia.Config.get()
 
-    # Determine if show is completed/ended
+    # Determine if show is completed/ended.
+    #
+    # Read through MetadataAccess: `metadata` is loaded as a %MediaMetadata{}
+    # struct with atom keys, so the string-keyed match this used to do could
+    # never succeed. `is_completed` was therefore always false and
+    # completed_show_refresh_threshold_hours was never read — every ended show
+    # was throttled as if it were still airing.
     is_completed =
-      case media_item.metadata do
-        %{"status" => status} when is_binary(status) ->
+      case MetadataAccess.get(media_item.metadata, :status) do
+        status when is_binary(status) ->
           String.downcase(status) in ["ended", "canceled", "cancelled"]
 
         _ ->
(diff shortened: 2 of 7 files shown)


arsfeld:
Addressed the review feedback in e09bd6a8.

**CodeRabbit's comment** (call-count assertion in `media_test.exs`): done. The handler now counts invocations and the test asserts exactly two TMDB show fetches, so a third cannot slip in unnoticed. Dropping to a bare `expect` would have lost that coverage.

**A separate review** challenged the previous commit's claim that the new cost was bounded by `should_skip_season_refresh?/1`. That was right, and following it up found the throttle could never fire at all, for two reasons beyond the one this PR started with:

- `seasons_refreshed_at` was missing from `MediaItem.changeset/2`'s cast list. Both writers go through `Media.update_media_item/3`, so every write was silently dropped and the column stayed NULL forever. That is why all 54 shows on the affected instance had it NULL, and a nil timestamp makes the throttle return false.
- The ended-show branch matched a string key against an atom-keyed `%MediaMetadata{}`, so `is_completed` was always false and `completed_show_refresh_threshold_hours` was never read.

Both fixed, each with a test verified failing beforehand. The comment now states the cost honestly rather than claiming a bound that did not exist: the throttle does not skip an airing show on a weekly pass, which is the intent.

Same review also caught that the new test stubbed a TVDB endpoint with TMDB-shaped keys, so the refresh wrote `year: nil` over the fixture while the assertions still passed. Now uses the real TVDB shape and asserts the year survives.

arsfeld:
Both comments were right, addressed in 4c953434.

**Mass-assignment of `seasons_refreshed_at`:** agreed, and it was worse than a style point. The field gates `should_skip_season_refresh?/1`, so making it castable let any `update_media_item/3` caller extend its own throttle window from attrs. It also contradicted CLAUDE.md's rule against casting programmatically-set fields, and the sibling `last_upgrade_check_at` immediately below it, which carries a comment saying exactly that and is written via `Repo.update_all`.

Followed that existing pattern rather than adding a changeset: dropped it from the cast list and added `Media.stamp_seasons_refreshed/1`, mirroring `Upgrades.stamp_checked/2`. `ProviderSwitch` no longer passes the field through `provider_switch_attrs` (where it was a silent no-op anyway) and stamps after the episode recreation clears its integrity check, inside the transaction so a rollback takes the stamp with it.

**Stamping after a partial failure:** also right. Season failures are logged and the reduce continues, so a partial pass looked complete. That was harmless while the write was silently dropped, but now that it persists it would throttle the retry and hide the missed season until the threshold expired. Now counts failed seasons and only stamps on a clean pass.

Three tests added, each verified failing first: the stamp persists, the field stays unwritable through `update_media_item/3`, and a 500 on one season leaves the timestamp nil.

Local: 7803 tests 0 failures, credo clean, format clean, compile clean under `--warnings-as-errors`.

arsfeld:
Both halves check out, addressed in 5eb8fb5d.

**Per-episode failures not reaching `failed_seasons`:** correct. `upsert_episodes_from_season/3` swallows per-episode changeset errors and still returns `{:ok, count}`, so a season could persist fewer episodes than the provider returned and still look clean. `create_episodes_for_season/3` now compares the upserted count against the episodes that were actually upsertable and returns `{:error, {:incomplete_episode_upsert, _, _}}` on a short write, which feeds the existing failed-season count. Kept it contained to `media.ex` rather than changing the shared upsert's contract across its four callers: the enricher and scanner have their own `create_episodes_for_season`, so nothing else shifts.

**Partial season selections stamping:** also correct, and live rather than theoretical. `season_monitoring` reaches `refresh_episodes_for_tv_show/2` from the add-media UI (`add_media_live/index.ex:226`), so `"first"`, `"latest"` and `"none"` are real values — and `"none"` fetches nothing at all yet still stamped. The timestamp means every season is current, so only a clean pass over the full set writes it now.

Test added for the partial selection; the short-write path is covered by the existing failed-season test's assertion that a non-clean pass leaves the timestamp nil.

Local: 7804 tests 0 failures, credo clean, format clean, compile clean under `--warnings-as-errors`.

PR #462, opened 2026-08-15
feat(cast): subtitles when casting to a Chromecast, over direct and p2p routes

arsfeld:
Subtitles now work when casting to a Chromecast, over both the direct-server route and the p2p/LAN bridge. The bridge could not serve subtitles at all before this.

## How it works

Both of Mydia's serving paths already hand out any path-validated file inside an HLS streaming session's temp directory: `HlsController` over HTTP, and `Mydia.P2p.Server` over the bridge. So a WebVTT file written into that directory is reachable on both transports with no new endpoint, no p2p protocol change, and no new authentication, since session auth already covers it.

`Mydia.Streaming.SessionSubtitles` materializes deliverable tracks as `subs_<trackId>.vtt` on demand, single-flighted. Extraction is lazy, and `Subtitles.Delivery` already caches embedded extractions on disk, so ffmpeg runs at most once per media file. Image-based tracks (PGS, VobSub) never materialize, enforced server-side so no client can offer one by mistake.

On the client, `CastRouteResolver` rewrites tracks into per-route URLs and puts them on the route, so the session manager reads `route.subtitles` and never needs to know which transport it got.

## Behaviour

Subtitles default to off, matching local playback, which never explicitly auto-selects. Casting from the player carries your current selection across. The cast bar has a caption picker; dismissing it leaves the receiver alone, which is distinct from choosing Off. The choice survives a seek restart, and a reconnect no longer wipes the stored record.

`dart_cast` unconditionally activates the first track it is given and offers no way to say "offer these, activate none". Two workarounds handle that: ordering the list so the intended track is first, and an explicit disable after load. Both are commented as workarounds where they appear.

## Requires a server that serves session-relative subtitle files

The direct Chromecast route now uses `/api/v1/hls/<session>/subs_<trackId>.vtt` instead of `/api/player/v1/subtitles/file/:fileId/:trackId`. A player on this version talking to an older server gets a 404 and shows no subtitles on that route, with no error surfaced.

This was a deliberate call. The alternative was keeping two URL shapes so the change stayed strictly additive. Uniformity won, so the constraint needs saying out loud: update the server alongside the player. The requirement is documented at the call site in `cast_route_resolver.dart`, not only here.

There is no GraphQL schema change, so nothing else in this PR has a version requirement.

## Security fixes picked up along the way

`HlsController` had three functions building paths from raw path parameters with no traversal validation: `root_segment/2`, `segment/2`, and `variant_playlist/2`. All three now route through `SessionFiles.safe_path/2`.

The last one is reachable rather than theoretical. Phoenix URI-decodes path segments before matching, so `GET /api/v1/hls/<sid>/..%2F..%2Fetc/index.m3u8` arrives with `track_id` decoded and reads outside the session directory. Its fixed `index.m3u8` filename narrows exploitation but does not prevent it.

`SessionFiles.safe_path/2` also fixes two subtler holes the old p2p validator had: it rejects an absolute name up front, because `Path.join/2` strips a leading `/` rather than treating it as an override, and it compares against the base plus a separator rather than a bare string prefix, which previously let a sibling session directory through on a shared prefix.

## Known gaps

- **Unverified:** whether a Chromecast fetches a track's body on activation or prefetches every track at LOAD. Lazy extraction assumes the former. If it prefetches, every cast pays an ffmpeg extraction per track inside a request process. The remedy is a background task at session start and changes no interface. This needs a real receiver to settle.
- Cold restart and stale-session reconnect do not repopulate the track list, so the receiver will not re-show a subtitle on those two paths. The stored choice is protected from corruption; it just has nothing to attach to. Pre-existing, documented in the manager.
- Casting while in direct play offers sidecar subtitles only, not embedded ones. In direct play the player rebuilds its track list from the playback engine, whose ids are engine-local and have no server-side identity, so those tracks cannot be addressed by the session route. This is unchanged from before the PR, where a `url != null` filter excluded them for the same underlying reason. Mapping engine track ids back to ffprobe stream indexes would fix it, but the two orderings are not guaranteed to agree and a wrong mapping shows the wrong subtitle rather than none.
- `Mydia.P2p.Server.handle_hls_session_stream/3` has no automated coverage. Its subtitle logic is four lines identical to the HTTP path, which is covered end to end.
- `find_session_by_id/2` ignores its `user_id` argument, so session lookup is by session id alone. Pre-existing and out of scope here, but worth its own issue: a leaked session id already yields the full video stream through the same lookup.

## Testing

`test/mydia_web/controllers/api/hls_controller_test.exs` is new. There were no HLS controller tests in the repo before, which is why `Mydia.Streaming.HlsSessionStub` exists: standing up a real session spawns ffmpeg, and `get_info/1` is a plain `GenServer.call`, so a process double at that seam exercises the real routing and auth without one.

Elixir: 7827 tests green including Credo. Flutter: the cast, player-screen and widget suites green at `--concurrency=1`, whole-project analyze clean.


arsfeld:
Both findings were valid. Fixed in 7fdc4f4d and ed0fde04.

**Atomic write.** Correct, and the fast path is exactly where it bites: `ensure/2` checks `File.exists?/1` before taking the lock, so a reader could observe an entry that `File.write/2` had only started creating. The body now lands under a unique temporary name and is published with a single rename.

One caveat worth stating rather than leaving implied. The two tests added alongside it do not discriminate this fix from the old direct write: both still pass if the change is reverted. Forcing the interleaving deterministically would mean instrumenting the module purely so a test could observe it, which is a worse trade than an honest gap. The fix is verified by inspection and by checking the rename and cleanup paths directly, not by a test that would catch a regression here.

**`mk_` ids.** Also correct, and it was a regression introduced two commits earlier in this same PR. The filter used to be `deliverable && url != null`, and the `url != null` half was removed so a freshly downloaded sidecar (which carries a UUID id but no URL) could be cast. What that half had also been doing, invisibly, was excluding direct-play tracks: in direct play the player rebuilds its list from the playback engine with ids like `mk_0` and no URL.

Filtering on the URL was always a proxy for the real question, so the fix filters on the id shape the server can actually serve, which is what `SessionSubtitles` accepts: a stream index or a UUID. Downloaded sidecars keep working, `mk_` ids are excluded, and the `deliverable` half is untouched since image tracks must never reach a receiver. The test breaks the filter two ways to confirm it fails without the fix.

Not mapping `mk_<n>` back to a stream index deliberately: the engine's track order and ffprobe's are not guaranteed to agree, and a wrong mapping shows the wrong subtitle rather than none. The resulting limitation, that casting from direct play offers sidecars only, is unchanged from before this PR and is now written into the description.

PR #470, opened 2026-08-15
feat: server-declared minimum player version

arsfeld:
A Mydia server now declares the oldest player version it works with, and the player carries the oldest server version it works with. A mismatch in either direction shows a banner in the app shell instead of surfacing as an unexplained failure.

## Why both directions

Server newer than player happens when an operator upgrades the container ahead of their phone. Player newer than server is the more common case in practice: mobile and desktop players update themselves from GitHub releases while a self-hosted server stays on whatever the operator last pulled, and an old server has no way to warn about it. So each side declares its own floor. The two are different facts and neither derives from the other.

## Transport: GraphQL, not `/health`

`/health` was the obvious choice and it was wrong. `native/mydia_p2p_core/src/lib.rs` defines the p2p request set as a closed enum (`Pairing`, `ReadMedia`, `GraphQL`, `HlsStream`, `Custom`) with no general HTTP tunnel. In p2p mode only GraphQL and HLS traverse the link, so a `GET {serverUrl}/health` goes out over the ordinary network to a URL that is typically unreachable, which is why the session is on p2p at all. That would have covered LAN while silently yielding nothing for every remote session, and remote users are the ones most likely to run a stale player.

Adding a `ServerInfo` variant to the p2p enum was rejected on a sharper ground: it is a serde enum, so an old server hitting an unknown variant fails to deserialize. That ships a cross-version protocol break inside the feature whose job is warning about cross-version breaks.

## What changed

**Server.** `Mydia.Compatibility` holds the two floors with the bump rule documented. A new public `serverCompatibility` root field publishes them alongside the server's own version. Public because a player that cannot authenticate is exactly the one that may need to be told why, and it exposes nothing `/health` does not already serve unauthenticated. Mirrored in the Rust `mydia-api` crate for the `sdl_parity` gate.

**Player.** A pure `evaluateCompatibility` function returns one of six verdicts. Required outranks recommended, and within a tier the player side wins, because the person holding the phone can act on that message themselves. The banner sits in `AppShell` above both layout branches and re-asks on app resume.

**Two tiers.** Required is persistent and non-dismissible. Recommended is dismissible, keyed on the exact version pair so it returns when either side moves. Required verdicts neither read nor write a dismissal key, so a stale key can never suppress a hard warning.

**Fail open throughout.** A failed query, a null root field, an unparseable version, or a pre-feature server that rejects the document all converge on `unknown`, which renders nothing. A bug in this feature degrades to silence, never to a false banner.

**Cleanup.** Deletes the dead four-layer `ProtocolVersion` scaffolding from the relay architecture iroh replaced, on both sides, plus its unreachable `UpdateRequired` dialog and the orphaned `PLAYER_UPDATE_URL` config whose only consumer was the deleted module. 664 lines.

## Baseline

All four floors ship at `0.9.0`, below the current `v0.13.2` release line, so nothing fires on this release. The floors are unreachable at ship time regardless of value, because only builds carrying this feature ever send or answer the query. The first real bump comes with the first breaking change after this lands.

## One defect worth calling out

Unversioned builds report `0.0.0-dev` (`mix.exs` falls back to it when `BUILD_VERSION` is unset, and `ci-docker.yml` does not pass it). That string *parses* cleanly to `0.0.0` and sorts below every floor, so an earlier revision of this branch raised a permanent non-dismissible "this server is out of date" banner against the `:master` image, every self-built image, and every local dev server. The fail-open guard covered versions that fail to parse, not versions that parse low. `evaluateCompatibility` now treats a `0.0.0` core on either side as unversioned rather than ancient, with a counter-test proving a genuinely old `0.8.0` server still warns.

## Cross-version behaviour

Two player versions will be in the wild at once and operators upgrade on their own timeline, so there is no deploy order here.

- New player, old server: the query fails validation, verdict is `unknown`, no banner. Same as today.
- Old player, new server: the old player never sends the query, so the field is inert. Same as today.
- Both new: works in both directions.

## Verification

- `mix precommit`: 7912 tests, 0 failures
- `flutter test --concurrency=1`: 2336 tests, 0 failures
- `cargo test -p mydia-api`: 47 tests across 7 binaries, 0 failures, including `sdl_parity` and `types_remaining`
- `mix schema.export`: no diff, SDL is current
- `mix mydia.graphql validate`: 36/36 operation files valid
- `flutter analyze`: 56 issues, exactly the pre-change baseline, zero introduced

## Follow-ups, deliberately not in this PR

- `ci-docker.yml` passes `DATABASE_TYPE` and `BUILD_COMMIT` but not `BUILD_VERSION`, so `:master` images report `0.0.0-dev`. The guard above makes that harmless, but stamping a real version would be better.
- `release.yml`'s `docker-build` job does not rewrite `player/pubspec.yaml`, so the web player bundled into released server images reports `1.0.0` rather than the release version. Benign until the release line crosses 1.0.0, at which point a bundled web player would tell the user to update the app that served it.
- A Hive box failure currently suppresses a recommended nudge; treating a box failure as "not dismissed" and still showing it would be the safer default.

## Summary by CodeRabbit

* **New Features**
  * Added compatibility checks between player and server versions.
  * Displays banners for required or recommended player and server updates.
  * Added a details view showing current and required versions, with links to Settings.
  * Recommended warnings can be dismissed for specific version pairings.
  * Compatibility status refreshes when the app resumes.
  * Added a public server compatibility query for supported version information.

* **Bug Fixes**
  * Pairing failures now display as standard connection errors.


arsfeld:
Both CodeRabbit findings are addressed in 5bbd453f.

**Major, the unguarded state write:** correct and reachable. `dismiss()` is called fire-and-forget from the banner's close button, so a disposal during the Hive write would have thrown `UnmountedRefException` as an unhandled async error. Added an `ref.mounted` guard before the assignment.

**Minor, the Hive fault suppressing the banner:** also correct, and it matches a finding from our own whole-branch review that we had deferred as a follow-up. Two independent reads agreeing moved it into scope. `build()`'s dismissal lookup is now caught and falls back to "not dismissed", so a storage fault keeps the warning visible instead of silently hiding it. The class docstring was updated to match.

Both paths have regression tests: a failing dismissal box must still show a recommended banner, and dismissal after disposal must not throw. Compatibility suite is 36/36, analyze unchanged at the 56-issue baseline.

arsfeld:
@coderabbitai review

PR #472, opened 2026-08-16
feat(import): make media import incremental, resumable, and stoppable

arsfeld:
## Why

Importing an existing library was all or nothing and required the browser tab to stay open for the whole run. All progress lived in the LiveView socket, so closing the tab, a disconnect, or a deploy lost everything. Combined with slow metadata matching, a large collection could not be imported at all.

This makes the import incremental and resumable. You can stop it at any point and keep everything processed so far, which lets you work a large library in as many sittings as you want instead of committing to one unbroken run.

## What changed

**The run lives in Oban, not in the page.** `Mydia.Jobs.ImportRun` coordinates on its own `imports: 1` queue, so a multi-hour import cannot starve library, music, or book scans on the shared `:media` queue. Closing the tab does nothing to the run; reopening re-attaches by reading the run row and subscribing to its topic.

**Two phases.** Phase 1 walks the tree and commits a `media_file` row per file, in batches of 100, with no HTTP at all. After it, every file on disk is durably recorded, so stopping during a scan loses nothing and a relay outage cannot destroy scan progress. Phase 2 matches outstanding files in chunks, caches a candidate per file, and in unattended mode links confident matches as it goes.

**There is no resume cursor.** Phase 1 commits rows and phase 2 commits links and candidates, so Continue is just a fresh run over the same path: existing rows are skipped, linked files are excluded, and files that already carry a candidate skip the relay entirely. The database is the cursor. This is what makes stopping at an arbitrary point safe rather than merely convenient, and it removes a whole class of bookkeeping bugs.

**Stopping is cooperative.** Stop sets `:stopping`; the coordinator re-reads its own row between chunks and drains. Nothing is rolled back, because every unit of work was committed as it completed.

**Matching got substantially cheaper.** `Mydia.Metadata.Cache` is a plain ETS get-or-compute with no single flight, so a season matched concurrently had every worker miss the cache at the same instant and issue the same search. `Library.BatchMatcher` groups a batch by normalized title, matches one file first to warm the cache, then fans out the rest. Twelve episodes of a season now cost one relay search instead of ten. Groups still run concurrently, so throughput is unchanged.

**A shared ingest core.** `Library.FileIngest` holds the per-file match, decide, commit step with two policies. `:local_only` reproduces what the scheduled scan has always done and links only a match that came from the local database. `:create_items` links anything at or above the confidence threshold. `Jobs.LibraryScanner` now delegates to it, so there is one implementation instead of two.

**The wizard is replaced by a permanent inbox.** `/import` is now a run control plus a review inbox over orphaned `media_file` rows: paged, filterable, with per-file editing keyed by `media_file_id` and batch editing. The four step wizard, its `import_sessions` table, its cleanup job, and its components are gone.

## Behaviour changes worth knowing

**The scheduled library scan now populates the inbox.** Routing it through `FileIngest` means it writes a match candidate for every file it cannot link, which it never did before. On your first cron scan after upgrading, orphaned files that have been invisible for months will appear in the review inbox. This is intended. The scan's linking decision is unchanged: it still never creates a new item from an external provider match.

**An interrupted import no longer auto-resumes.** Previously a stale job sweep re-queued an orphaned coordinator about an hour after the interruption. Now a run interrupted by a restart is reconciled to a terminal state at boot and its job is retired, so restarting is an explicit Start click. The new contract is better: recovery is immediate rather than up to an hour of a dead spinner, and it covers imports interrupted inside the first hour, which the old path never did. No work is lost, because the database is the cursor.

**The orphan health check narrowed.** `DatabaseHealthCheck` no longer counts files that are queued for import review. Without this, a large first scan after upgrading would light the health check red for doing nothing wrong. It still reports genuinely stuck orphans.

## Migration

Two new tables (`media_file_match_candidates`, `import_runs`) and one drop (`import_sessions`). All portable across SQLite and PostgreSQL. The drop discards any in-flight wizard session; those expired after 24 hours anyway and rescanning rebuilds the state. `down/0` reconstructs the original table faithfully, constraints and indexes included.

## Testing

7900+ tests green on the full `mix precommit` gate. Everything is verified on both SQLite and PostgreSQL, including the cross-adapter cases that bite in this codebase: NULL sort position in `ORDER BY`, constraint name reporting differences between the two Ecto adapters, and join cardinality.

The properties the feature is sold on have tests written to fail against a plausible broken implementation, several proven by deliberately breaking the code and confirming the right test went red:

- A stop after a committed batch retains that batch, and a resumed run finishes the remainder.
- A resumed run issues no new relay calls for files that already carry a candidate.
- Twelve episodes of a season produce exactly one relay search.
- The inbox editor edits the file you clicked, not the row at that position, across paging and filter changes.
- A boot-interrupted run is released so its library path is not locked.

## Summary by CodeRabbit

* **New Features**
  * Added import runs with progress tracking, start/stop controls, recovery, and history.
  * Added an inbox for filtering, editing, approving, and batch-processing unmatched files.
  * Added metadata matching, confidence indicators, saved candidates, and manual movie or TV episode matching.
  * Added unattended imports with automatic high-confidence linking.
* **Bug Fixes**
  * Interrupted imports are detected and reconciled after restart.
  * Unmatched files and library-type mismatches now show clearer statuses and errors.
* **Refactor**
  * Replaced import sessions with persistent import runs and inbox-based review.


arsfeld:
All six review findings are addressed in 92075a52. The comments above were re-anchored to new line numbers on the follow-up commit rather than re-raised, so here is where each one landed:

- **Client-controlled parameters.** `mode` and `filter` now map through explicit allow-lists (`@modes`, `@filters`) via `Map.fetch/2`, and `offset` parses safely with a non-negative guard. Stored `media_type` is validated on the approve path too.
- **Batch-edit logic.** Moved to `Library.apply_batch_match/3`, which runs the reads and writes in one transaction and returns updated and failed counts. The flash now reports failures instead of claiming success for every selected file. This also removes a per-file read-plus-write round trip that was 200 queries for a full page selection.
- **Core components.** The run control now uses `<.input type="select">` and `<.button>`. DOM ids are unchanged so the existing tests still target them.
- **`Task.async_stream` failure handling.** Switched to `Task.Supervisor.async_stream_nolink/4` against `Mydia.TaskSupervisor`, handling `{:exit, reason}` alongside `{:ok, _}`. Kept `ordered: true` specifically so an exit can be zipped back to its originating path: the coordinator's termination invariant depends on every input path yielding exactly one result, so a crashed file has to become an error result rather than vanish.
- **Unresolvable paths.** Dropped ids are now logged. Worth noting the log was initially invisible: `config/config.exs` carries a Logger metadata allow-list and `:media_file_ids` was not on it, so the ids were stripped in production as well. Added there too.
- **`file_ingest`.** Fixed; the concurrent `LibraryScanner`/`ImportRun` race on the rank-0 unique index is handled by widening the error path that `scan_result_from_ingest/1` already covers.

Twelve new tests, each confirmed to fail against the pre-fix code first.

Two deliberate deviations, both commented in place:

- `inbox.ex` row location uses `MediaFile.display_path/1` rather than `display_name/1`. That line sits under a title that already falls back to the filename, so `display_name/1` would render the basename twice for every unidentified file, which is the common case in this queue.
- `#batch-toggle-<id>` stays a bare checkbox. It is a `phx-click` target rather than a form field, and `<.input type="checkbox">` would add a dead hidden input and break the row layout.

One behaviour change worth a second look during review, also called out in the description: an interrupted import no longer auto-resumes. Previously the stale-job sweep re-queued an orphaned coordinator about an hour later. Now the run is reconciled to a terminal state at boot and its job retired, so restarting is an explicit action. Recovery is immediate rather than up to an hour of a dead spinner, and it covers interruptions inside the first hour that the old path never did. No work is lost, because the database is the cursor.

PR #487, opened 2026-08-18
fix(show): keep every in-flight add's spinner on the recommendations rail

### LINKED ISSUE #459: Recommendations rail loses a card's spinner when two adds are in flight
On the movie and show detail page, adding two titles from the "More like this" rail at once makes one card's spinner vanish while its add is still running, and a re-click on that card is then silently swallowed.

## Where

`MydiaWeb.MediaLive.Show.RecommendationEvents` tracks in-flight adds correctly, in a `MapSet` under `:adding_recommendation_tmdb_ids`. But it also maintains a single `:adding_recommendation_id`, and that is the only thing the rail compares against to decide whether to draw a spinner. `clear_in_flight/2` picks an arbitrary survivor:

```elixir
|> assign(
  :adding_recommendation_id,
  remaining |> Enum.at(0) |> then(&if(&1, do: to_string(&1)))
)
```

The module's own comment concedes the split:

> The MapSet is the source of truth for "is this id already in flight" and is what prevents a double-click from racing two adds. The single id alongside it is only what the card compares against to show its spinner.

So with two adds running, when the first finishes, `Enum.at(0)` may hand back the wrong id. The still-running card loses its spinner and reads as idle. Clicking it again is correctly dropped by the `MapSet.member?/2` guard in `dispatch_add/2`, so nothing breaks, but from the user's side the card looks stuck and unresponsive.

## Why this is now easy

#457 added a per-item `:adding` key to `DiscoverComponents.media_rail/1` for exactly this reason, because a franchise routinely has several adds in flight. The franchise adapter uses it (`MapSet.member?(adding_tmdb_ids, entry.tmdb_id)`); the recommendations rail was left on the old single-id path.

## Fix

Have `RecommendationEvents.decorate/2` (or the render path) set `adding` per item from the existing `MapSet`, and stop passing `adding_item_id` at `show.html.heex`. The `:adding_recommendation_id` assign can then be deleted entirely.

Found during review of #457.

arsfeld:
Adding two titles at once from the "More like this" rail made one card's spinner vanish while its add was still running, and a re-click on that card did nothing visible.

Closes #459.

## Root cause

One step earlier than the issue describes. The issue points at `clear_in_flight/2` picking an arbitrary survivor with `Enum.at(remaining, 0)`, which is real but only bites with three or more adds in flight.

The primary defect was `mark_in_flight/2`. It overwrote `:adding_recommendation_id` with the newest id, and that single id was the only thing the card compared against to decide whether to draw a spinner. So the first card lost its spinner the moment Add was clicked on a second card, before anything had finished. The re-click was then correctly dropped by the `MapSet.member?/2` guard in `dispatch_add/2`, which is what made the card look stuck rather than merely idle.

Both symptoms are the same error: a single id cannot represent a set.

## What changed

- `RecommendationEvents.with_in_flight/2` stamps the per-item `:adding` flag that `trending_card/1` already honours (`discover_components.ex:160`), computed from the MapSet at render. It runs at render rather than inside `decorate/3` because the set changes on every add and every result, while `decorate/3` runs once when the lookup lands.
- The MapSet is now the only in-flight state. The `:adding_recommendation_id` assign, both of its writes, and `adding_item_id` at the call site are deleted. `clear_in_flight/2` is a plain `MapSet.delete`, so the arbitrary-survivor logic is gone rather than corrected.

This is the pattern the franchise strip has used since #457 (`franchise_components.ex:69`). The recommendations rail was left on the old single-id path.

## Verification

`./dev mix precommit` green on all seven steps: 8168 tests, 0 failures.

Three new tests, written before the fix and confirmed failing on the missing function:

- a two-element in-flight set flags both items and leaves a third alone, which is the issue's exact scenario
- an empty set flags nothing
- an unparseable `provider_id` is treated as not in flight rather than raising, since `safe_provider_id/1` returns nil there

The wiring itself has no dedicated test. Driving two real adds through a connected LiveView would need heavy stubbing of the add path for little gain, and the existing connected rail tests already render through the new template expression, so a wrong arity or assign name fails them.

### DIFF
--- a/lib/mydia_web/live/media_live/show.ex
+++ b/lib/mydia_web/live/media_live/show.ex
@@ -141,7 +141,6 @@ defmodule MydiaWeb.MediaLive.Show do
      |> assign(:recommendations, [])
      |> assign(:recommendations_expanded, false)
      |> assign(:adding_recommendation_tmdb_ids, MapSet.new())
-     |> assign(:adding_recommendation_id, nil)
      |> assign(:requesting_recommendation_id, nil)
      |> assign_new(:metadata_config, fn -> Mydia.Metadata.default_relay_config() end)
      |> assign(
--- a/lib/mydia_web/live/media_live/show.html.heex
+++ b/lib/mydia_web/live/media_live/show.html.heex
@@ -207,10 +207,14 @@
             collapsible={@media_item.type == "tv_show"}
             expanded={@recommendations_expanded}
             toggle_event="toggle_recommendations"
-            items={@recommendations}
+            items={
+              RecommendationEvents.with_in_flight(
+                @recommendations,
+                @adding_recommendation_tmdb_ids
+              )
+            }
             media_type={if @media_item.type == "tv_show", do: :tv_show, else: :movie}
             current_user={@current_user}
-            adding_item_id={@adding_recommendation_id}
             requesting_item_id={@requesting_recommendation_id}
             can_add={@can_create_media}
             on_select={nil}
--- a/lib/mydia_web/live/media_live/show/recommendation_events.ex
+++ b/lib/mydia_web/live/media_live/show/recommendation_events.ex
@@ -211,6 +211,20 @@ defmodule MydiaWeb.MediaLive.Show.RecommendationEvents do
      |> put_flash(:error, "Could not add that title")}
   end
 
+  @doc """
+  Stamps each item's `:adding` flag from the set of in-flight adds.
+
+  Applied at render rather than inside `decorate/3`, because the set changes on
+  every add and every result while `decorate/3` runs once, when the lookup lands.
+  A single id cannot describe more than one in-flight add, which is what #459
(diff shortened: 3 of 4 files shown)


arsfeld:
Superseded by #485, which landed while this was in CI and closed #459.

That fix is broader than this one. It replaces both `adding_item_id` and the per-item `:adding` key with a single `adding_ids` MapSet on `trending_card/1` and `media_rail/1`, converts all four hosts rather than just the recommendations rail, and normalises membership across the integer/string boundary in `adding?/1`. This PR only rewired the recommendations rail and left the dual mechanism in place elsewhere.

Nothing here is worth salvaging on top of it. The edge case this PR tested separately, an unparseable `provider_id`, cannot raise under #485's string comparison, and its two-concurrent-spinner assertion is covered by #485's own rail test.

Closing as duplicate work rather than resolving the conflict.

Issue #278, opened 2026-07-31
Postgres: file-size column is int4 and overflows on files larger than ~2.1 GB


arsfeld:
On PostgreSQL, an integer column is `int4` but receives byte-size values larger than 2147483647, so every write of a file over ~2.1 GB fails.

```
Postgrex expected an integer in -2147483648..2147483647, got 7565057364.
```

Observed values include `7565057364` and `8324361099`, which are plainly file sizes in bytes (~7.5 GB and ~8.3 GB).

Crash site: `lib/postgrex/type_module.ex:1084` in `Postgrex.DefaultTypes.encode_params/3`, raised as `DBConnection.EncodeError`.

### Evidence (relay crash telemetry, as of 2026-07-31)

| | |
|---|---|
| Occurrences | **127,571** |
| Distinct instances | 1 (Postgres user) |
| Version | 0.12.0 |
| First / last seen | 2026-06-26 / 2026-07-31 (ongoing) |

One instance is in a permanent crash loop on this. It accounts for roughly **99% of all crash reports the relay has ever received**, which drowns out every other signal in the crash dashboard.

### Suggested fix

Find the size column(s) still typed `:integer` and migrate to `:bigint` on Postgres. SQLite stores INTEGER as 64-bit so it silently works there, which is why this is Postgres-only and was not caught. Worth auditing every byte-size / duration column in one pass rather than fixing only the one that is currently crashing.

PR #505, opened 2026-08-19
fix(downloads): stop a download deleted mid-flight from crashing jobs and LiveViews

### LINKED ISSUE #281: Ecto.NoResultsError: jobs use Repo.one! on a download that can be deleted mid-flight
A job loads a download with `Repo.one!` and raises when the row is gone by the time it runs.

```
Ecto.NoResultsError: expected at least one result but got none in query:
  from d0 in Mydia.Downloads.Download,
  where: d0.id == ^"...",
  preload: [:media_item]
```

Crash site: `lib/ecto/repo/queryable.ex:173` via `Ecto.Repo.Queryable.one!/3`.

### Evidence (relay crash telemetry, as of 2026-07-31)

| | |
|---|---|
| Occurrences | 356 (plus 5 more from a second call site at `queryable.ex:178`) |
| Distinct instances | 2 (plus 1) |
| Versions | 0.12.0 and dev builds |
| First / last seen | 2026-06-26 / 2026-07-31 (ongoing) |

### Suggested fix

This is an ordinary race: the download is deleted or cleaned up between the job being enqueued and it running. Use `Repo.one/1` (or `get/2`) and treat `nil` as "download no longer exists, nothing to do" rather than crashing the job.

arsfeld:
Fixes #281.

## What was happening

`DownloadMonitor.perform/1` reads every download row from the database, *then* polls each configured client over HTTP, then fans the result out over six `Enum.each` passes that each reload their row by id. In production seconds pass between the snapshot and the reload — long enough for an operator delete, a concurrent import, `Queue`/`Grabber` cleanup, or the monitor's own reject path to remove the row.

Five of those six handlers used the raising `get_download!/2` with no guard, so a row deleted inside that window raised `Ecto.NoResultsError` and aborted the whole job. Every handler queued behind the failing one was silently skipped for that poll. That is the same amplification #278 fixed for the stall-detection reduce, in the same file, in the passes it did not cover.

356 occurrences across two instances between 2026-06-26 and 2026-07-31, ongoing. The reported query's `preload: [:media_item]` is a usable fingerprint: it matches `handle_completion`, `handle_failure`, `reject_bad_content` and `handle_missing` exactly, and among all other call sites in the repo only `downloads_live/index.ex`'s `match_files_import`.

## The root cause was the missing primitive

`Mydia.Downloads` exposed only `get_download!/2`. Nil-safety was therefore opt-in: six call sites hand-rolled `rescue Ecto.NoResultsError` around the bang getter and five forgot, with nothing in the signature to say which was which. Sprinkling a sixth rescue would have left the next call site exactly as exposed.

So this adds `get_download/2` — the same shape as the existing `Library.get_media_file/2` and `Collections.get_collection_by_id/2` — and converts every one of the 43 call sites in `lib/`, deleting the ad-hoc rescues as it goes.

## The write half was broken too

`Repo.update/2` and `Repo.delete/2` filter by primary key on every write and raise `Ecto.StaleEntryError` when that matches zero rows. This is the default for any write, **not** something `optimistic_lock` opts you into, and the non-bang variants raise just the same. So `update_download/2` and `delete_download/1` raised if the row vanished between a caller's reload and its write — an exposure wider than #281, covering every `update_download` caller including `handle_stale_grab/1`, which writes a struct from `list_stale_grabs/1` with no reload at all.

Fixed at the choke point with Ecto's own `:stale_error_field`, which converts the raise into the `{:error, changeset}` every caller already handles. Four lines in one file; no call site needed changing. Two strict `{:ok, _} =` matches that would have become `MatchError` got real error branches.

## LiveViews

All 20 download reads in the web layer sit in `handle_event` handlers — none in `mount` or `handle_params`, where raising a 404 would be correct. So every one was a crash-the-socket bug: delete a download in one tab, click it in another, and the second tab's LiveView process died. Each module gets a `with_download` helper that flashes and re-renders instead, closing any modal that was about the now-deleted row.

The two batch handlers differ, since they iterate a selection: `batch_delete` counts an already-deleted row as removed (the operator asked for it gone and it is gone), while `batch_retry` does not count it (nothing left to retry).

## Tests

The race reproduction is deterministic and needs no sleeps. `list_downloads_with_status/1` reads the rows before it calls any client, so deleting a row **inside the Bypass handler that answers the client poll** lands in the exact window on every run.

I verified the test earns its keep by reverting `DownloadMonitor` to the raising reload: both race tests fail with `Ecto.NoResultsError`, and pass again with the fix. The load-bearing assertion is that the *other* download in the same batch was still handled — asserting only that `perform/1` returns `:ok` would pass even if the poll bailed out early.

Also adds a guard test asserting `get_download!` appears nowhere under `lib/`. The function stays for tests, where a row that vanished unexpectedly should fail loudly; it just may not be called from application code. Without it this fix erodes the same way the bug arrived.

## Verification

- `mix test test/mydia/downloads_test.exs test/mydia/downloads/ test/mydia/jobs/` — 1628 tests, 0 failures
- `mix test test/mydia_web/live/downloads_live/ test/mydia_web/live/media_live/` — 334 tests, 1 failure
- `mix precommit`

The one web failure is `ManualSearchInfoLinkTest`, which is unrelated and pre-existing: it fails in isolation at `--seed 0` with this branch's `lib/` changes reverted to master's, and a different manual-search test fails at other seeds. Nothing in this diff touches manual search.

## Note for review

`evaluate_and_apply/3` keeps its own containment rather than being folded into the new `with_download/4`. It evaluates the in-memory struct before touching the database (`StallDetector.evaluate/7` can raise first), reloads only inside some decision branches, and must return a counted `{stalled, failures}` tuple rather than `:ok`. Its now-unreachable `Ecto.NoResultsError` clause is gone; the broad rescue and the `catch :exit` from #278 stay. There is a comment saying so, so nobody force-unifies the two later.

## Summary by CodeRabbit

* **Bug Fixes**
  * Download actions now handle records deleted during processing without crashing.
  * Monitoring, retrying, deleting, importing, matching, and rejecting downloads continue safely when downloads disappear.
  * Background transfers stop normally instead of retrying deleted downloads.
  * Stale dialogs and actions refresh appropriately and display a clear “That download no longer exists.” message.
  * Concurrent updates and deletions return recoverable errors instead of raising unexpected exceptions.

* **Tests**
  * Added regression coverage for deleted downloads across monitoring, background processing, batch actions, and LiveView interactions.

### DIFF
--- a/lib/mydia/downloads.ex
+++ b/lib/mydia/downloads.ex
@@ -122,6 +122,15 @@ defmodule Mydia.Downloads do
   defdelegate get_download(id, opts \\ []), to: Mydia.Downloads.History
 
   @doc """
+  Whether a changeset from a download write failed because the row was gone,
+  rather than because the write was invalid.
+
+  See `Mydia.Downloads.History.stale_changeset?/1`.
+  """
+  @spec stale_changeset?(Ecto.Changeset.t()) :: boolean()
+  defdelegate stale_changeset?(changeset), to: Mydia.Downloads.History
+
+  @doc """
   Gets a single download.
 
   ## Options
--- a/lib/mydia/downloads.ex
+++ b/lib/mydia/downloads.ex
@@ -107,12 +107,31 @@ defmodule Mydia.Downloads do
   defdelegate list_downloads_with_status(opts \\ []), to: Mydia.Downloads.History
 
   @doc """
+  Gets a single download, or `nil` if it no longer exists.
+
+  ## Options
+    - `:preload` - List of associations to preload
+
+  Prefer this over `get_download!/2` anywhere the row can be deleted underneath
+  you: background jobs, client pollers, PubSub handlers, and any LiveView event
+  handler acting on an id the browser sent. A download disappearing between the
+  read that produced its id and the read that acts on it is an ordinary race, not
+  an exceptional condition (issue #281).
+  """
+  @spec get_download(binary(), keyword()) :: Download.t() | nil
+  defdelegate get_download(id, opts \\ []), to: Mydia.Downloads.History
+
+  @doc """
   Gets a single download.
 
   ## Options
     - `:preload` - List of associations to preload
 
   Raises `Ecto.NoResultsError` if the download does not exist.
+
+  This is for tests and assertions, where a missing row should fail loudly.
+  Application code under `lib/` must use `get_download/2` instead — enforced by
+  `Mydia.Downloads.NoRaisingDownloadGetterTest`.
   """
   @spec get_download!(binary(), keyword()) :: Download.t()
   defdelegate get_download!(id, opts \\ []), to: Mydia.Downloads.History
--- a/lib/mydia/downloads/client/debrid/fetcher.ex
+++ b/lib/mydia/downloads/client/debrid/fetcher.ex
@@ -267,9 +267,18 @@ defmodule Mydia.Downloads.Client.Debrid.Fetcher do
           descriptor
       end)
 
-    download = History.get_download!(state.download_id)
-    new_metadata = Map.merge(download.metadata || %{}, %{"debrid_urls" => persisted})
-    History.update_download(download, %{metadata: new_metadata})
+    case History.get_download(state.download_id) do
+      nil ->
+        # Cancelled or deleted while we were resolving. Report it as an ordinary
+        # fetch failure so the caller's error path runs, rather than raising out
+        # of the GenServer and bypassing both the retry back-off and
+        # `fail_download/2` (issue #281).
+        {:error, Error.api_error("download row no longer exists")}
+
+      download ->
+        new_metadata = Map.merge(download.metadata || %{}, %{"debrid_urls" => persisted})
+        History.update_download(download, %{metadata: new_metadata})
+    end
   end
 
   defp stream_all(urls, download_dir, state, download) do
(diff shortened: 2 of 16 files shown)


arsfeld:
@coderabbitai review

arsfeld:
Thanks — the Major finding was a real regression I introduced, addressed in 43f073b67.

**1. Fetchers retrying a deleted row (Major) — valid, fixed.**

Confirmed against the code: `handle_info/2` in both fetchers retries every `{:error, _}` while the budget lasts, and each retry re-runs `run/1` from the top. So returning an error for a missing row meant the debrid fetcher makes a fresh provider call to re-resolve URLs, and from `finalize/2` — which runs *after* the whole payload has transferred — re-downloads all of it, three times over.

The seedbox case is worse than described: `run/1` calls `maybe_delete_remote/2` *before* `finalize/1`, so a retry re-opens SFTP for a remote path it has already deleted.

Both now return a terminal `:download_deleted` that `handle_info/2` matches ahead of the retry clause and stops on. There is nothing left to fetch for and no row to mark failed, so stopping is the whole correct behaviour.

Both regression tests assert the fetcher exits inside a 2s budget against a 5s first-retry delay, which is what actually distinguishes "terminal" from "retried". I verified the debrid one fails against the pre-fix code with `fetcher did not exit in time`.

**2. Guard test excluding the definition files (Minor) — valid, fixed.**

Correct that `raising_calls/1` matches a *qualified* `.get_download!(`, which `def`/`defdelegate`/`@spec` cannot trip, so the exclusion only blinded the guard to a real call added to those files later. The exclusion is gone and the guard still passes.

One note: the committable suggestion also dropped the `|> Path.wildcard()` line, which would have left a glob string being read as a file path and broken the test. Applied by hand instead, removing only the `Enum.reject`.

**3. Asserting the modal closes (Minor) — valid, fixed.**

Right that the flash alone would pass with the modal left open on a download that no longer exists. Now asserts `#match-files-modal` is present after opening and absent after the submit.

arsfeld:
Valid, fixed in 6dad3b643.

Confirmed against the code: the previous commit made a nil *lookup* terminal but left the narrower window open. If the row is deleted between that lookup and the write, `update_download/2` now returns a stale changeset (from the `:stale_error_field` added earlier in this PR), and that fell straight through to the generic `{:error, _}` retry path — the same expensive retry storm, just via a smaller gap.

Rather than pattern-matching a stale changeset in three places, this adds `Downloads.stale_changeset?/1` next to the `@stale_opts` that create the condition, so the two halves stay together. It reads Ecto's `stale: true` error tag (set by `Ecto.Repo.Schema.apply/4`) rather than the message text, so rewording `@stale_opts` can't silently break it. All three write sites map only that result to `:download_deleted`; an ordinary changeset error stays retryable, which is the distinction that matters.

**On the requested tests.** I put them at the context level, where the race is genuinely deterministic: delete the row, write, assert the predicate — plus a validation-error case so it's proven to be more specific than "any changeset error".

I did not add fetcher-level tests for this particular window. To hit it you'd have to delete the row between `get_download/2` and `update_download/2` *inside* the GenServer, and there's no seam there — adding one that exists only for the test would be inventing production structure to prove a predicate that's already directly tested. The fetchers keep their lookup-window tests, which do run end-to-end through `handle_info/2` and assert the process stops inside a 2s budget against a 5s retry delay.

`mix precommit` is running now; I'll confirm it green before merging.

arsfeld:
@coderabbitai review

PR #506, opened 2026-08-19
fix(downloads): tolerate a download row deleted mid-flight

### LINKED ISSUE #281: Ecto.NoResultsError: jobs use Repo.one! on a download that can be deleted mid-flight

### LINKED ISSUE #285: Ecto.StaleEntryError when updating a Download concurrently
Updating a `Download` raises `Ecto.StaleEntryError`, meaning the row changed or was deleted between load and update.

```
Ecto.StaleEntryError: attempted to update a stale struct:
  %Mydia.Downloads.Download{__meta__: #Ecto.Schema.Metadata<:loaded, "downloads">, ...}
```

Crash site: `lib/ecto/repo/schema.ex:1023` via `Ecto.Repo.Schema.apply/4`.

### Evidence (relay crash telemetry, as of 2026-07-31)

| | |
|---|---|
| Occurrences | 21 |
| Distinct instances | 1 |
| Version | 0.12.0 |
| First / last seen | 2026-06-27 / 2026-07-30 |

Likely the same underlying race as the `Repo.one!` crash: the download monitor and the import/queue paths both touch a download concurrently, and one deletes or updates it while the other holds a stale struct.

### Suggested fix

Handle the stale case explicitly (`Repo.update/2` returning `{:error, :stale}` via `stale_error_field`, or re-fetch and retry). Worth looking at alongside the `Repo.one!` issue since they point at the same concurrency model around downloads.

arsfeld:
Closes #285. Closes #281.

Two crash reports turned out to be the same bug seen from two sides: a `downloads` row is deleted by one process while another still holds a struct for it.

`Mydia.Downloads.Download` has no `optimistic_lock` and no `lock_version`, so Ecto's `UPDATE ... WHERE id = $1` can never fail because of a concurrent *field* update. On this table `Ecto.StaleEntryError` has exactly one meaning: **the row is gone** — the same condition `Ecto.NoResultsError` reports on the read side. That is why both issues share call sites and are fixed together at one choke point rather than at the ~40 call sites.

### Where it actually crashed

Recovered from the relay's crash store as full stacktraces, not inferred:

| Crash | Call site | Hits |
|---|---|---|
| NoResultsError | `download_monitor.ex handle_completion/1` | 382 + 7 |
| NoResultsError | `download_monitor.ex handle_failure/1` | 66 + 2 |
| NoResultsError | `media_import.ex` import path | 85 |
| StaleEntryError (update) | `media_import.ex do_process_import` | 9 |
| StaleEntryError (update) | `download_monitor.ex update_progress/2` | 7 |
| StaleEntryError (update) | `download_monitor.ex handle_completion/1` | 6 |
| StaleEntryError (delete) | `download_monitor.ex handle_failure/1` | 6 |
| StaleEntryError (delete) | `queue.ex create_download_record_with_retry/4` | 1 |
| NoResultsError | Downloads / media LiveViews | 5 |

Rows vanish because `DownloadMonitor.perform/1` iterates a snapshot taken before minutes of client polling, while a second monitor pass (the adaptive fast-followup deliberately is not a uniqueness duplicate of the cron tick), `MediaImport`, or the operator deletes rows underneath it. Preventing that overlap was considered and rejected: an operator deleting from the UI causes the identical race, so the writes have to tolerate it regardless.

### The fix

Both halves are off-the-shelf Ecto options, so behaviour is identical on SQLite and Postgres (staleness is decided by rows-affected in shared `Ecto.Adapters.SQL` code).

In `Mydia.Downloads.History`:

- `update_download/2`, `mark_download_completed/1`, `mark_download_failed/2` pass `stale_error_field: :id`. A vanished row now returns the ordinary `{:error, changeset}` shape every caller already handles instead of raising, and new `stale_error?/1` distinguishes it from a validation failure. Keeping it a real changeset matters: ~15 callers do `inspect(changeset.errors)` in their error branch, so returning a bare atom would have traded `StaleEntryError` for `UndefinedFunctionError`.
- `delete_download/1` passes `allow_stale: true` and is now **idempotent** — a row another process already deleted still returns `{:ok, download}` and still broadcasts, so open LiveViews drop it. This removes the delete-stale crash class with no change at any of the 16 delete call sites, including one that hard-matches `{:ok, _} =`.
- New non-raising `get_download/2`, the sibling of `get_download!/2`.

Call sites whose *behaviour* had to change, not just their logging:

- `handle_completion/1` no longer hard-matches `{:ok, _} = mark_download_completed/1`, and skips the completion event and import job when the row is gone.
- The stall path reports a vanished row at debug rather than counting it in `stall_update_failures`, leaving `evaluate_and_apply/3`'s broad rescue for genuine bugs.
- `MediaImport` treats a vanished row after a successful import as success — the files are already on disk, and crashing there made Oban retry and import them twice.
- `MediaRematch` keeps a correct, disk-verified relink instead of rolling it back over an audit-only provenance stamp it can no longer write.
- `attach_upgrade_target/2` no longer reports a grab that actually succeeded as a failed initiate, which could provoke a duplicate grab.
- `Grabber` skips conflict resolution when its own row is the missing one, rather than deleting an unrelated row on its behalf.
- The debrid and seedbox fetchers stop cleanly when a download is cancelled mid-fetch instead of crashing the GenServer.
- LiveView handlers say the download is no longer in the queue and re-render instead of crashing.

Three `rescue`-based one-offs (`MediaImport`'s `Ecto.StaleEntryError` guard and two `Ecto.NoResultsError` fetch helpers) are folded into the new return values, so there is one mechanism rather than several.

### Testing

- Choke-point contract covered directly: stale update returns a `stale: true` changeset, a validation error is *not* misflagged, delete is idempotent and still broadcasts, `get_download/2` returns `nil`.
- Job-level regressions for the three confirmed crash sites, driven by a telemetry handler that deletes the row at the poll's primary-key read. It matches on the SQL rather than counting `downloads` queries — `perform/1` issues several and their order shifts with fixtures, so a counter silently tests the wrong window. Each asserts the injected delete actually fired, so they cannot pass without exercising the race. One asserts the rest of the poll still runs, and one asserts the file still reaches the library.
- Reverting `lib/` makes 11 of the new tests fail; with it, `mix test` is 8446 tests / 0 failures. `mix format --check-formatted` and `mix credo --strict` are clean.

The remaining `get_download!/2` call sites are all inside existing rescues with correct semantics (the stall path, the batch loops, `reject_bad_content/1`) and were left alone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling when downloads are deleted during processing, transfers, imports, or status updates.
  - Prevented unnecessary retries, errors, and failure notifications for downloads that no longer exist.
  - Download actions now show an informational message and refresh the list when a record is unavailable.
  - Deletion is safely repeatable and remains synchronized across the interface.
  - Improved resilience when download updates become outdated during concurrent changes.

- **Tests**
  - Added coverage for concurrent deletion scenarios across fetching, monitoring, and media imports.

### DIFF
--- a/lib/mydia/downloads.ex
+++ b/lib/mydia/downloads.ex
@@ -118,6 +118,24 @@ defmodule Mydia.Downloads do
   defdelegate get_download!(id, opts \\ []), to: Mydia.Downloads.History
 
   @doc """
+  Gets a single download, or `nil` if it no longer exists.
+
+  Prefer this over `get_download!/2` in background jobs and LiveView handlers,
+  which act on an id captured earlier and must tolerate the row having been
+  imported, cleared, or deleted in the meantime (issue #281).
+  """
+  @spec get_download(binary(), keyword()) :: Download.t() | nil
+  defdelegate get_download(id, opts \\ []), to: Mydia.Downloads.History
+
+  @doc """
+  True when a write failed because the download row no longer exists.
+
+  See `Mydia.Downloads.History.stale_error?/1`.
+  """
+  @spec stale_error?(term()) :: boolean()
+  defdelegate stale_error?(changeset), to: Mydia.Downloads.History
+
+  @doc """
   Creates a download.
   """
   @spec create_download(map()) :: {:ok, Download.t()} | {:error, Ecto.Changeset.t()}
@@ -125,6 +143,9 @@ defmodule Mydia.Downloads do
 
   @doc """
   Updates a download.
+
+  Returns `{:error, changeset}` if the row no longer exists; `stale_error?/1`
+  distinguishes that from a validation failure.
   """
   @spec update_download(Download.t(), map()) :: {:ok, Download.t()} | {:error, Ecto.Changeset.t()}
   defdelegate update_download(download, attrs), to: Mydia.Downloads.History
@@ -148,6 +169,8 @@ defmodule Mydia.Downloads do
 
   @doc """
   Marks a download as completed by storing the completion time.
+
+  Returns `{:error, changeset}` if the row no longer exists; see `stale_error?/1`.
   """
   @spec mark_download_completed(Download.t()) ::
           {:ok, Download.t()} | {:error, Ecto.Changeset.t()}
@@ -155,6 +178,8 @@ defmodule Mydia.Downloads do
 
   @doc """
   Records an error message for a download.
+
+  Returns `{:error, changeset}` if the row no longer exists; see `stale_error?/1`.
   """
   @spec mark_download_failed(Download.t(), String.t()) ::
           {:ok, Download.t()} | {:error, Ecto.Changeset.t()}
@@ -162,6 +187,8 @@ defmodule Mydia.Downloads do
 
   @doc """
   Deletes a download.
+
+  Idempotent: a row another process already deleted still returns `{:ok, download}`.
   """
   @spec delete_download(Download.t()) :: {:ok, Download.t()} | {:error, Ecto.Changeset.t()}
   defdelegate delete_download(download), to: Mydia.Downloads.History
--- a/lib/mydia/downloads/client/debrid/fetcher.ex
+++ b/lib/mydia/downloads/client/debrid/fetcher.ex
@@ -149,6 +149,15 @@ defmodule Mydia.Downloads.Client.Debrid.Fetcher do
       {:ok, new_state} ->
         {:stop, :normal, new_state}
 
+      # The download was cancelled while this fetcher was working. Terminal, not
+      # retryable: there is nothing left to fetch for (issue #281).
+      :deleted ->
+        Logger.info(
+          "Debrid fetch stopping: download_id=#{state.download_id} was removed mid-fetch"
+        )
+
+        {:stop, :normal, state}
+
       {:error, reason} when state.retries_left > 0 ->
         Logger.warning(
           "Debrid fetcher attempt failed for download_id=#{state.download_id} " <>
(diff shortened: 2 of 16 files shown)


arsfeld:
@coderabbitai review

arsfeld:
@coderabbitai review

arsfeld:
@coderabbitai review

arsfeld:
Thanks — four of the five are fixed in 044fca2d3:

- **Stale changesets in the retry / retry-import / keep-waiting handlers** — correct, they showed a generic update error where the row had actually vanished. They now go through `download_gone/1` like the read-side paths.
- **Match modal state in `download_gone/1`** — correct, `match_files_import` and `submit_match` reach it from an open modal, so the dialog survived its own row. Cleared now.
- **`dismiss_failed_grab/2` swallowing delete failures** — correct, and a regression I introduced: replacing the old hard `{:ok, _} =` match with the idempotent call turned a genuine failure into a silent reload. A real error flashes now.
- **The `refute_received` in the fetcher test** — good catch, that assertion could not fail. The stub runs in Bypass's handler process, so `send(self(), ...)` never reached the test. The pid is captured outside now.

**Not taking the staged-files nitpick**, for three reasons:

1. **It is not introduced by this PR.** Before this change the same branch called `get_download!`, raised `Ecto.NoResultsError` and crashed the fetcher — leaving exactly the same bytes in staging. The `:deleted` path makes the stop graceful; it does not create the leak.
2. **Deleting those files is a policy decision that already has an owner.** `Queue.cancel_download/2` threads `delete_files` through to `Client.remove_download/4`. A fetcher error path cannot see whether the operator asked for files to be kept, so unilaterally removing them there would override that flag.
3. **The severity argument points the other way.** The note says that with `delete_after_transfer` the seedbox path has already removed the remote source, so the local copy is the only remaining one. That is a reason not to delete it from an error path, not a reason to.

Happy to file a follow-up for a staging-directory sweep if orphaned staging bytes are worth reclaiming on their own — that seems like the right shape for it, rather than an unconditional `rm` on a race path.

arsfeld:
Superseded by #505, which merged first and covers the bulk of this — the choke-point change in `Mydia.Downloads.History`, the non-raising getter, the DownloadMonitor and MediaImport call sites, the LiveView handlers, and the fetchers' terminal deleted-row path.

Closing rather than rebasing: the overlap is large enough that the remaining diff is easier to review on its own.

The parts #505 did **not** cover are now in #508 — the `MediaRematch` rollback that discards a correct relink, `Grabber` resolving a unique-constraint conflict against its own missing row, and `attach_upgrade_target/2` reporting a successful grab as a failed initiate.

One deliberate difference worth recording: this branch made `delete_download/1` idempotent via Ecto's `allow_stale: true`, so a row another process already deleted returns `{:ok, download}`. #505 kept `stale_error_field` for delete and handles the error per call site instead. Both avoid the crash; #505's approach is what shipped, and #508 does not re-litigate it.

PR #507, opened 2026-08-19
fix(player): make sign out actually sign out on macOS

arsfeld:
On macOS, Sign out did nothing. The app stayed on the Settings screen, still signed in, with no error shown.

## Cause

`_handleSignOut` ran three awaits with the session-ending one last:

```dart
await ref.read(settingsServiceProvider).clearSettings();   // 1
await ref.read(connectionProvider.notifier).clear();       // 2
await ref.read(authStateProvider.notifier).logout();       // 3
```

`SettingsService.clearSettings()` fired four `FlutterSecureStorage.delete()` calls with no error handling, and two of those keys (`library_sort_movies`, `library_sort_tvShows`) are absent unless the user changed a library sort. On the macOS legacy keychain, deleting an absent key throws. `auth_storage_native.dart` already documents this: `NativeAuthStorage._withFallback` exists precisely because of it. `SettingsService` was the one credential store that bypassed that hardening.

So step 1 threw, steps 2 and 3 never ran, and the exception disappeared into a dropped `VoidCallback` Future. An unanswered macOS keychain prompt produces the same symptom by hanging instead.

## Also fixed: sign-out forgot almost nothing

`clearSession()` cleared four keys and left the entire pairing credential set in the keychain, including `pairing_device_token`. That token mints fresh access tokens through `refreshAccessToken`, which is deliberately unauthenticated on the server because the device token is itself the proof of identity. `PairingService.clearCredentials()` already existed to delete exactly those keys and was never called from anywhere.

## Changes

- `SettingsService` takes an injectable `AuthStorage` instead of holding a private `FlutterSecureStorage`, so its deletes degrade rather than throw.
- New `SessionTeardown` erases every server-scoped credential behind one storage instance: session, relay URL, twelve pairing keys, pinned certs, connection diagnostics, settings.
- New `bestEffort` guard wraps each step, swallowing failures and timing out hangs, so no single uncooperative keychain call can abort sign-out.
- `AuthStateNotifier.logout()` owns the whole sequence. It outlives the router redirect that unmounts the Settings screen, which is why the ordering the widget used to juggle no longer matters.
- `CertStorage` gains an injectable storage; `clearCredentials()` picks up the `pairing_media_token_expiry` it was missing.
- The Settings screen shrinks to a confirmation plus one call.

`device_id` is deliberately retained. It is this installation's stable identity, and regenerating it would make the next pairing create a second device row on the server and orphan the first. That is also why the wipe uses an explicit key list rather than `AuthStorage.deleteAll()`.

## Tests

2474 passing, 9 new. The tap path had no coverage at all before, which is how this shipped.

- `clearSettings()` completes against a real `NativeAuthStorage` wrapping a failing backend, which is the macOS bug reproduced.
- `SessionTeardown` erases every server-scoped key and leaves `device_id`, asserted by exact equality.
- Teardown continues when the keychain refuses a delete, and gives up on a step that never completes.
- `logout()` reaches `unauthenticated` even when a cleanup step throws, covering both the async path and a synchronous throw from `ref.read`.
- Tapping Sign out and confirming actually signs out; cancelling does not.

## Not included

Server-side `revokeDevice`. The mutation exists and the player still never calls it, so a signed-out device stays listed and trusted under Remote Access. Deliberately out of scope here.

## Verification gap

Manual macOS verification has not been done: the development host is Linux. The bug is macOS-only, so this wants a real run of sign out, quit, relaunch on a Mac before it ships to users.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive sign-out cleanup for session data, credentials, certificates, settings, and connection diagnostics.
  * Preserves the device identity during sign-out.
  * Added media-token expiry cleanup when pairing credentials are cleared.
  * Sign-out cleanup continues despite individual failures or timeouts.

* **Bug Fixes**
  * Ensured users reach the signed-out state even when cleanup errors occur.
  * Removed claim codes from pairing logs.

* **Tests**
  * Added coverage for cleanup, failure handling, credential removal, device ID preservation, and sign-out confirmation behavior.

### DIFF
--- a/player/lib/core/auth/cert_storage.dart
+++ b/player/lib/core/auth/cert_storage.dart
@@ -8,7 +8,11 @@ import 'auth_storage.dart';
 /// certificate fingerprints for each Mydia instance. This enables
 /// certificate pinning for self-signed certificates.
 class CertStorage {
-  final AuthStorage _storage = getAuthStorage();
+  /// [storage] is injectable for tests and for session teardown, which builds
+  /// every credential store over one storage instance.
+  CertStorage({AuthStorage? storage}) : _storage = storage ?? getAuthStorage();
+
+  final AuthStorage _storage;
 
   static const _fingerprintsKey = 'cert_fingerprints';
 
--- a/player/lib/core/auth/session_teardown.dart
+++ b/player/lib/core/auth/session_teardown.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/foundation.dart' show debugPrint;
 
 import '../channels/pairing_service.dart';
+import '../connection/connection_diagnostics_provider.dart';
 import '../settings/settings_service.dart';
 import 'auth_service.dart';
 import 'auth_storage.dart';
@@ -43,10 +44,12 @@ class SessionTeardown {
 
   final AuthStorage _storage;
 
-  /// Wipes the session, the pairing, the pinned certificates and the settings.
+  /// Wipes the session, the pairing, the pinned certificates, the settings
+  /// and the connection diagnostics.
   ///
-  /// Steps are independent. The session goes first because it is the wipe that
-  /// matters most, so it lands before anything else can misbehave.
+  /// Each step is wrapped in [bestEffort], so no step's outcome depends on
+  /// any other's and none can block the ones after it. The order below is a
+  /// priority ordering only; it provides no safety guarantee on its own.
   Future<void> run() async {
     final auth = AuthService(storage: _storage);
 
@@ -60,6 +63,10 @@ class SessionTeardown {
     await bestEffort(
       'settings',
       SettingsService(storage: _storage).clearSettings,
+    );
+    await bestEffort(
+      'diagnostics',
+      () => clearConnectionDiagnostics(_storage),
     );
   }
 }
--- a/player/lib/core/auth/session_teardown.dart
+++ b/player/lib/core/auth/session_teardown.dart
@@ -1,0 +1,65 @@
+import 'package:flutter/foundation.dart' show debugPrint;
+
+import '../channels/pairing_service.dart';
+import '../settings/settings_service.dart';
+import 'auth_service.dart';
+import 'auth_storage.dart';
+import 'cert_storage.dart';
+
+/// How long a single cleanup step may take before it is abandoned.
+///
+/// The macOS keychain can block on an access prompt that nobody is there to
+/// answer. Without a ceiling, that prompt keeps the user signed in.
+const _stepTimeout = Duration(seconds: 5);
+
+/// Runs [step], swallowing failure and refusing to wait forever.
+///
+/// Sign-out has to finish even when storage is uncooperative, so a step that
+/// throws or hangs is logged and skipped. The alternative is leaving the user
+/// signed in, which is the bug this exists to prevent.
+Future<void> bestEffort(
+  String label,
+  Future<void> Function() step, {
+  Duration timeout = _stepTimeout,
+}) async {
+  try {
+    await step().timeout(timeout);
+  } catch (e) {
+    debugPrint('[SessionTeardown] $label cleanup failed: $e');
+  }
+}
+
+/// Erases every credential this device holds for the server it was signed into.
+///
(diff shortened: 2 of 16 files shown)


arsfeld:
@coderabbitai review

arsfeld:
@coderabbitai review

arsfeld:
Good catch on the per-key isolation, and it was a real one. Fixed in 7a0e2d0.

`bestEffort` isolates failures between teardown steps but not within one, so the twelve sequential awaits in `clearCredentials()` meant any earlier failure stranded every later delete. `pairing_device_token` was sixth in that sequence, and it is the credential that still mints access tokens through the unauthenticated `refreshAccessToken` mutation, so the security consequence was exactly as described.

Verified rather than assumed: on native this was masked, since `NativeAuthStorage.delete` swallows through `_withFallback`. But that is a property of one implementation, not a guarantee of the `AuthStorage` interface, and the web implementation calls `localStorage.removeItem` unguarded, where a throw is live.

Both `clearCredentials()` and `clearConnectionDiagnostics()` now use `Future.wait`, matching the shape `SettingsService.clearSettings()` already used. Added a regression test that makes the *first* key's delete throw and asserts `pairing_device_token` is gone anyway; confirmed it fails against the old sequential code (`Expected: false / Actual: <true>` on the device-token check) and passes now.

Full suite 2475/2475.

arsfeld:
Addressed in 7e5aca7.

You were right that the mechanism is real, and I confirmed it reproduces: with the fix stashed and a non-`async` test double that throws before returning a Future, `pairing_device_token` survived the wipe (`Expected: false / Actual: <true>`). Both cleanup functions now wrap each deletion in `Future.sync`.

Worth recording for anyone reading this later: every `AuthStorage` implementation in the tree today (`NativeAuthStorage`, `_WebAuthStorage`, the stub, and the test mock) declares `delete` as `async`, and Dart guarantees an `async` function reports errors through its Future rather than throwing synchronously. So this was not a live bug. It is hardening: `AuthStorage` is an interface, and the wipe guarantee should not depend on every future implementer remembering to write `async`. That is the same reasoning behind the previous fix, so it would have been inconsistent to invoke it there and wave it off here.

Full suite 2476/2476.

arsfeld:
Also addressed the claim-code logging finding, in 5183374.

Verified provenance first: those two `debugPrint` lines are pre-existing and byte-identical in content at the merge base. They only show up in this diff because `dart format` rewrapped them when the hook touched this file earlier in the branch. So this PR surfaced the issue rather than introduced it.

Fixed anyway, since it is a live credential in device logs and the file was already open. Both sites now log `deviceName` and `platform` and omit the code, which keeps a pairing traceable without putting the thing that authenticates it into the log. Checked the rest of the file for the same shape: no other debug output interpolates a claim code, device token, access token, or media token.

Full suite 2476/2476, analyze unchanged at 58 pre-existing info lints.

arsfeld:
Respectfully pushing back on this one: I believe it conflates two different keys, and applying the proposed diff would be a regression.

There are two similarly-named keys:

- `_StorageKeys.deviceId` = `'pairing_device_id'` (`pairing_service.dart:66`). Server-assigned during pairing, part of the pairing credential set. Deleting it on sign-out is correct.
- `DeviceInfoService._deviceIdKey` = `'device_id'` (`device_info_service.dart:15`). This installation's stable identity, and the one the logout contract protects. `SessionTeardown` never touches it.

So `clearCredentials()` deleting `pairing_device_id` does not violate the contract; `device_id` is a different key and is preserved.

Two existing tests already pin both halves of this:

- `pairing_service_clear_test.dart`, 'clearCredentials leaves the device identity alone': seeds `device_id` and `pairing_device_token`, asserts only `device_id` remains.
- `session_teardown_test.dart:65`: seeds the full key set including `pairing_device_id` (line 16) and `device_id` (line 34), then asserts `storage.contents` equals exactly `{'device_id': 'stable-id'}`.

That second assertion is exact equality, so removing the `pairing_device_id` deletion as proposed would leave it behind and turn that test red. The suite would catch the change immediately.

Not applying this one. Happy to reconsider with evidence if I have misread something.

PR #512, opened 2026-08-20
ci(web): build the player for web, deploy it on release

arsfeld:
Closes the three gaps left open after web.mydia.dev went live. They share one failure mode: a break in the web bundle stayed invisible until someone deployed.

## 1. CI never built the player for web

`ci-player.yml` covers Windows, macOS, Linux and Android, and `icons` only diffs the generated SVG set. Nothing ran `flutter build web`. 183 commits landed on `player/` between the public web player merging (#370, 2026-08-09) and its first deploy without one of them being compiled against the web target.

The new **Build / Web** job runs the deploy's exact flags minus the wasm module. Skipping the Rust half is deliberate rather than lazy: it needs nix, wasm-pack and a pinned `flutter_rust_bridge` install (roughly twelve minutes, too much per player PR), and `ci.yml` already compiles `native/mydia_p2p_core` for `wasm32-unknown-unknown`. What stays uncovered is wasm-pack's packaging step, whose inputs are pinned toolchain versions that move only on a deliberate bump. Dart source, which every PR touches, is what this catches.

It also asserts `build/web/_headers` survived, mirroring `build_web.sh`. Without that file the deployed page is never cross-origin isolated, there is no SharedArrayBuffer, and `RustLib.init()` throws in every browser.

## 2. web.mydia.dev now ships on every stable release

`deploy-web-player.yml` gains a `workflow_call` entry point; `release.yml` calls it after `publish`. The public player now moves in step with the tag, the Docker `:latest` image and the docs.

Reused rather than restated, because the build is nine steps of nix, devenv, Cachix and a pinned codegen install, and a second copy would drift from the one exercised between releases.

Two details worth review:

- **Stable only**, matching the existing `docs` job. web.mydia.dev is a single public surface with no staging copy, so a prerelease published there reaches every viewer with no way to opt in. Betas stay on the installable builds, which a user chooses. Say the word if you would rather betas deploy to a preview URL.
- **`ref` carries the SHA the draft pins**, not master's tip, so the bundle, the tag and the images all describe one commit even after master moves. `branch` is passed explicitly as `master` because a release checks out a commit and has no branch name to infer; it must equal the `mydia-web` Pages project's production branch or the deploy goes green to a preview URL and leaves the site on the previous release. A bare dispatch still falls back to `github.ref_name`, so a feature branch publishes to a preview rather than overwriting the public player.

## 3. The service worker's 503 race

`sw.js` raises `REPLY_TIMEOUT_MS` from 30s to 60s. It was tied exactly with the instance's own HLS readiness budget (`lib/mydia/p2p/server.ex:454`), so a cold start that legitimately used its full budget lost the race and returned 503 on a working session. That 503 is indistinguishable from a genuine transport fault, which made it the most misleading failure the worker could produce. The headroom covers the instance's wait plus the relay round trip carrying the answer back, and a browser connection never hole-punches, so it always pays relay latency.

This was item 0 on `MANUAL-VERIFICATION.md`, flagged as worth fixing before anyone reads browser matrix results.

## Verification

- The new job's exact sequence (`pub get`, `build_runner build`, the web build, the `_headers` assertion) runs green locally. Build takes 51.8s on top of existing setup.
- `node --check player/web/sw.js` passes, and the built `build/web/sw.js` carries `60_000`. Worth checking by hand because Flutter copies `web/` across verbatim, so a syntax error there would not fail a build.
- `actionlint` reports no new findings across the three workflows. The remaining hits (`release.yml` 323-856, `ci-player.yml:84`) are present on master unchanged.
- Reusable-workflow wiring checked structurally: declared inputs and passed inputs match exactly, and `prepare` really does output `sha` and `is_prerelease`.

## Not covered

The browser matrix still has not run, so the wasm iroh transport, relay dial, pairing and GraphQL-over-p2p remain unexercised anywhere. Relay capacity is also still one replica at 256Mi / 500m, and a browser session never hole-punches. Both are tracked in `docs/superpowers/sdd-web-player/MANUAL-VERIFICATION.md` and neither is affected by this PR.

## Summary by CodeRabbit

- **New Features**
  - Stable releases now automatically deploy the web player to production.
  - Web player deployments support both preview and production release workflows.
  - Preview and release deployments can be initiated with specified source and target branches.

- **Bug Fixes**
  - Increased the web player’s page-loading timeout to 60 seconds, improving reliability during slower cold starts and streamed content initialization.

- **Chores**
  - Added automated web-player build verification to ensure release assets are generated and configured correctly.

### DIFF
--- a/.github/workflows/ci-player.yml
+++ b/.github/workflows/ci-player.yml
@@ -91,6 +91,73 @@ jobs:
           echo "Browser tests: $files"
           flutter test --platform chrome $files
 
+  # Nothing compiled the player for the web until this job existed. The matrix
+  # below covers Windows, macOS, Linux and Android, and `icons` only diffs the
+  # generated SVG set, so 183 commits landed on player/ between the public web
+  # player merging and its first deploy without one of them being checked
+  # against the web target. That gap is only visible at deploy time, where it
+  # would now break a release: deploy-web-player.yml runs from release.yml.
+  #
+  # This builds the same bundle the deploy builds, with the same flags, minus
+  # the wasm module. The Rust half needs nix, wasm-pack and a
+  # flutter_rust_bridge codegen install (player/tool/build_web.sh), which is
+  # roughly twelve minutes and too much for every player PR. The split is
+  # deliberate rather than lazy: ci.yml already compiles
+  # native/mydia_p2p_core for wasm32-unknown-unknown, so the crate reaching
+  # wasm is covered, and what stays uncovered here is wasm-pack's packaging
+  # step, whose inputs are pinned toolchain versions that move only on a
+  # deliberate bump. Dart source, which every PR touches, is what this catches.
+  #
+  # MYDIA_WEB_P2P=true matters even without the module present: it selects the
+  # public-web branch in main.dart, so this compiles the code path the deploy
+  # actually ships rather than the instance-hosted one.
+  build-web:
+    name: Build / Web
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        working-directory: player
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Flutter
+        uses: subosito/flutter-action@v2
+        with:
+          flutter-version-file: player/.fvmrc
+          channel: stable
+          cache: true
+
+      - name: Install dependencies
+        run: flutter pub get
+
+      - name: Run code generation
+        run: flutter pub run build_runner build
+
+      - name: Build the web bundle
+        run: |
+          flutter build web --release --base-href / \
+            --no-web-resources-cdn \
+            --pwa-strategy=none \
+            --dart-define=MYDIA_WEB_P2P=true
+
+      # Same assertion build_web.sh makes, moved forward to PR time. Without
+      # _headers the deployed page never becomes cross-origin isolated, there
+      # is no SharedArrayBuffer, and RustLib.init() throws in every browser.
+      # Flutter copies web/ across verbatim, so this fails only if that file
+      # is deleted or renamed, which is exactly the change worth catching in
+      # review rather than in a browser.
+      - name: Check _headers survived the build
+        run: |
+          [ -f build/web/_headers ] || {
+            echo "::error::_headers did not make it into build/web. The deployed page" >&2
+            echo "         would not be cross-origin isolated and the p2p module would" >&2
+            echo "         fail to instantiate in every browser." >&2
+            exit 1
+          }
+
   build:
     name: Build / ${{ matrix.platform }}
     strategy:
--- a/.github/workflows/ci-player.yml
+++ b/.github/workflows/ci-player.yml
@@ -115,6 +115,13 @@ jobs:
     name: Build / Web
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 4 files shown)


arsfeld:
Thanks, all three were valid. Addressed in d70c470dc.

**`secrets: inherit` → named secrets.** The strongest of the three. It handed this workflow every secret the repo holds (Apple and Android signing keys, the Flatpak GPG key, R2 credentials) to build a static bundle that needs two tokens. Both are now named at the call site and declared under `workflow_call`. The same names resolve to repository secrets on a bare dispatch, so the steps read `secrets.*` unchanged on both paths. It was also the only `secrets: inherit` in the repo, so tightening it fights no convention.

**`_headers` contents, not just existence.** Agreed, and it mattered more than it looks: the file surviving the copy was never the likely failure. An edit removing the COOP or COEP line breaks the deployed page exactly as badly and would have passed. It now greps for both directives. Verified both ways: green against a real build, and red with the COEP line stripped out, so the guard demonstrably fails when it should.

**`contents: read` + `persist-credentials: false` on `build-web`.** Added. This job runs third-party code through `flutter pub get` and `build_runner`, so it is the one place in this file where a token in `.git/config` has something to bite on. Checked first that it is safe: `pubspec.yaml` has no git-sourced dependencies and the job runs no git commands.

Two deliberate limits on that last one, worth flagging rather than leaving you to wonder:

- Scoped to the new job instead of swept across `ci-player.yml`. The existing jobs have no `permissions:` blocks at all, and retrofitting them is its own change with its own blast radius.
- Not applied to `deploy-web-player.yml`'s checkout. That job enters a `devenv` shell which installs git hooks on entry, and it is the publishing path. `contents: read` is set there, but I left credential persistence alone rather than risk the one workflow that ships to production on an unverified assumption.

arsfeld:
On the out-of-diff finding for `ci-player.yml:150-155` (assert `build/web/pkg/mydia_player_p2p.js`): declining this one, because applying the proposed diff would fail the job on every run.

`web/pkg/` is produced solely by the wasm half of `player/tool/build_web.sh` (wasm-pack plus `flutter_rust_bridge build-web`), and it is gitignored (`player/.gitignore:129`). The `Build / Web` job deliberately does not run that half, which the comment above the job explains: it needs nix, wasm-pack and a pinned `flutter_rust_bridge_codegen` install, roughly twelve minutes, which is too much for every player PR. Verified on this branch, running the job's exact command:

```
$ ls player/build/web/pkg
ls: cannot access 'player/build/web/pkg': No such file or directory
```

So the module is never present here, and asserting it would be a guaranteed red rather than a guard.

The premise behind the finding is already covered where it belongs. `build_web.sh:150` carries that exact assertion, and that is the script the deploy runs, so a deploy cannot ship a bundle missing the module. The split is intentional and now stated in the commit message and the job comment: `ci.yml` compiles `native/mydia_p2p_core` for `wasm32-unknown-unknown`, so the crate reaching wasm is covered on every PR; what remains uncovered between releases is wasm-pack's packaging step, whose inputs are pinned toolchain versions that move only on a deliberate bump.

Worth noting the deploy path is not theoretical: it ran green end to end earlier today and the live bundle serves a 4.57 MB module with a shared `WebAssembly.Memory`, so the assertion in `build_web.sh` is doing its job.

PR #514, opened 2026-08-20
feat(player): rebuild the up-next prompt on the OSD design system

arsfeld:
The next-episode prompt was the only piece of player chrome not built on the OSD design system. It was a flat `Colors.black @ 0.7` box with Material `FilledButton`/`OutlinedButton` actions, `Theme.textTheme` typography, and a fixed 320px width pinned at `right: 24, bottom: 120` — floating over frosted, tier-aware, token-driven glass a few pixels away.

This rebuilds it as a two-state control in the OSD's own material: a resting `GlassPill` that expands into a glass card. The pill is deliberately the same object `SkipSegmentButton` just vacated in the same corner, so credits start, the skip pill appears, then the up-next pill takes its place. It also removes the mobile special case rather than papering over it — a pill fits at 360px, a 320px card never did.

## Four defects fixed along the way

- **The thumbnail never rendered.** `player_screen.dart` never passed `thumbnailUrl`, so every viewer got the grey play-icon placeholder. The data was already fetched; `season_episodes.graphql` selects the field.
- **The countdown ring divided by a hardcoded `10`**, under the comment "Assuming 10-second countdown", silently coupled to `_autoPlayCountdownDuration`. The ring now reads a `fraction` and never learns the total.
- **`SkipSegmentButton` took its geometry from the outlier.** It documented its insets as "Matches `UpNextOverlay`" — the widget following the design system deriving from the one ignoring it. Both now read `PanelMetrics.cornerInsetBottom`, guarded by a test that asserts the corner overlay never intersects the panel at any tier width.
- **A prompt could appear and then play nothing.** `_maybeShowUpNext` never checked that the next episode had files, while `_playNextEpisode` read `files.first` and returned silently when the list was empty — so the prompt could drain to zero and do nothing at all, with no feedback. `UpNextTarget` cannot be constructed without a file id, so a prompt that appears is now one that can be fulfilled.

## Stopping it is fast and reliable

Auto-play is the only thing in the player that acts without being asked, so the way out carries more weight than the way in. Ordered by when each layer can act:

- **Before the clock starts** — the dismiss lives on the resting pill, not behind the expand. A dismiss that costs one interaction to reveal and another to confirm is not a fast stop. `×` sits at the trailing edge behind a hairline with a 44px hit target, so a thumb reaching for it cannot catch Play.
- **While it runs** — presence holds the countdown. It does not advance while the pointer is over the prompt, it holds focus, or it is expanded. Reaching for the control is itself enough to stop the clock. On touch this yields a stop for free: tapping to expand holds it until an explicit choice.
- **While it runs** — Escape and a backward seek cancel. Escape already returned `handled` unconditionally, so the prompt takes the first branch and fullscreen-exit keeps the second. Someone scrubbing back into the episode is plainly not finished with it.
- **At the moment it fires** — a race guard. Auto-play never fires within 1s of the last input event; if the countdown lands inside that window it re-arms. This makes "I was reaching for Dismiss and it went anyway" impossible rather than merely unlikely. The timer cancels synchronously and `_playNextEpisode` re-checks the cancelled flag before navigating.

Dismissal stays sticky for the file. That is structural rather than a preference: `_maybeShowUpNext` runs on every position tick and `shouldOfferUpNext` stays true across the whole credits segment, so a non-sticky dismiss would re-show the prompt on the next frame.

## Season crossing

A finale now offers the next season's premiere — the lowest `episodeNumber >= 1`, so a special numbered 0 is never mistaken for it, and only when that episode has a file. An empty result or a failed fetch caches its outcome so a missing season is not refetched several times a second off the position tick.

This required a real fix: `_navigateToEpisode` hardcoded `seasonNumber=${widget.seasonNumber}` into the route. Crossing a season with that unchanged would have loaded the premiere while telling the new `PlayerScreen` it was still in the season it just left, so its `_fetchSeasonEpisodes` would fetch the wrong list, `_currentEpisodeIndex` would resolve to -1, and up-next would be dead for that entire season.

## Structure

The 256-line widget splits into three units under `video_controls/`, each testable without a `Player`, a `GraphQLClient`, or a widget tree:

| File | Holds |
| --- | --- |
| `up_next_policy.dart` | When to offer and what to offer: `shouldOfferUpNext`, `UpNextCandidate`, `UpNextTarget`, the two resolvers |
| `up_next_countdown.dart` | Timer, composable holds, race guard, injectable clock |
| `up_next_prompt.dart` | Resting pill, expanded card, engagement reporting |

The resolvers deliberately take `UpNextCandidate` rather than the generated GraphQL type, which is what lets them be unit tested without codegen having run; `player_screen` owns the one-line adaptation.

## Notes for review

- **All card text is `white @ 0.80` with `GlassPill.textShadow`.** This is measured, not aesthetic: `pill_legibility_test.dart` found this material puts unshadowed `white @ 0.80` text at 2.47:1, under the 4.5:1 WCAG SC 1.4.3 floor, and at 7.76:1 with the shadow. `up_next_prompt_legibility_test.dart` guards that conclusion by walking every `Text` in the card and failing with the offending string.
- **The countdown ring is white, not the brand amber.** The player chrome carries no hue anywhere; a coloured ring would have made the new widget the one exception all over again.
- `_upNextPlayingSub` has four teardown sites: `_cancelAutoPlay`, `_playNextEpisode`, `_playPreviousEpisode`, and `dispose`.
- Out of scope by explicit decision: undo-after-jump, a settings toggle for auto-play, and a series-finale terminal state.

## Verification

- `./dev flutter test --concurrency=1` → **2522 passed**, 0 failures. The flag is load-bearing; without it the suite silently drops files while still reporting success.
- `./dev flutter analyze` → 0 errors.
- No `--no-verify` on any commit; the pre-commit hook ran `dart analyze` and `dart format` clean throughout.

## Summary by CodeRabbit

* **New Features**
  * Added a redesigned Up Next experience with countdowns, pause and engagement handling, dismiss controls, episode artwork, and season-aware navigation.
  * Added lazy loading for next-season episodes and improved keyboard, touch, and responsive interactions.
  * Improved responsive placement and sizing of player controls across screen sizes.

* **Bug Fixes**
  * Seeking backward now dismisses Up Next.
  * Countdown prompts cancel reliably during navigation, media changes, and playback pauses.
  * Improved control visibility updates after touch interactions.


arsfeld:
@coderabbitai review

PR #517, opened 2026-08-20
Only offer season orderings TVDB publishes

arsfeld:
The TV show page offered all three of `SeasonOrder.values/0` for every TVDB show, with no knowledge of which orderings that show actually has. Most shows only publish an official ordering, so two of the three options were guaranteed failures: picking one landed on `{:error, :no_alternative_ordering}` and the flash "TVDB does not offer a DVD order ordering for this show."

The page now looks up the show's published orderings once on the connected mount, in the same async task that already fetches the suggestion banner's counts, and the selector renders only those. A show with a single ordering, or one whose lookup fails, renders no selector at all.

- `SeasonOrder.available/2` returns the orderings TVDB publishes, narrowed to what the `season_order` enum can store, always including the one the show is currently in
- The mount-time lookup now runs for every TVDB show the viewer can edit rather than only oversized ones, and returns availability and the banner counts together from one task, since `Metadata.Cache.fetch/3` has no locking and two tasks would both miss on a cold cache
- The stale-cache flash is reworded from "does not offer a DVD order ordering" to "does not list DVD order"

A show sitting on an ordering TVDB no longer publishes keeps that option and stays switchable back to aired order, so this cannot strand anyone.

## Testing

- 5 new unit tests for `available/2`, seeded through `Metadata.Cache` so they make no HTTP request and stay `async: true`
- 4 new and 2 rewritten LiveView tests, including a pair that guards against a hard-coded "drop absolute" shortcut passing for the wrong reason
- Full suite green (8521 tests), `--warnings-as-errors` clean, Credo clean
- A `--trace` run of the season-order UI file confirms all 23 tests finish under 700ms, so none reach the live relay

## Summary by CodeRabbit

* **New Features**
  * Added season-order options based on supported TVDB orderings.
  * Added a selector for switching between available season-order options.
  * Preserved the current season ordering when available options are displayed.
  * Combined season-order information and alternative-order suggestions into one loading flow.

* **Bug Fixes**
  * Unsupported and invalid ordering types are now excluded.
  * Improved handling for missing TVDB information and unavailable ordering data.
  * Added safeguards for stale submissions and unauthorized changes.

* **Tests**
  * Expanded coverage for ordering availability, filtering, authorization, and UI behavior.

### DIFF
--- a/lib/mydia/media/season_order.ex
+++ b/lib/mydia/media/season_order.ex
@@ -77,6 +77,43 @@ defmodule Mydia.Media.SeasonOrder do
   def effective(%MediaItem{season_order: order}), do: order
 
   @doc """
+  The orderings this show can actually be put into: the ones TVDB publishes
+  for it, narrowed to the ones `season_order` can store, plus whichever one
+  the show is in right now.
+
+  Meant to drive the ordering selector. Offering an ordering TVDB does not
+  publish buys a guaranteed `{:error, :no_alternative_ordering}` from
+  `switch/3`, since `Relay.fetch_ordering_episodes/4` finds no season stubs
+  of that type and returns an empty list. Not offering it is the fix.
+
+  `effective/1` is included unconditionally. A show recorded as `:dvd` whose
+  TVDB payload no longer lists a dvd ordering would otherwise render a
+  selector with no matching option, which misreports where the show is and
+  leaves no way back.
+
+  TVDB publishes types outside `@values` for some series ("alternate",
+  "regional"). They are dropped: `media_items.season_order` is an
+  `Ecto.Enum` restricted to `@values` and cannot store them.
+  """
+  @spec available(MediaItem.t(), map()) :: {:ok, [atom()]} | {:error, term()}
+  def available(%MediaItem{tvdb_id: nil}, _config), do: {:error, :missing_tvdb_id}
+
+  def available(%MediaItem{} = media_item, config) do
+    provider_id = to_string(media_item.tvdb_id)
+    current = effective(media_item)
+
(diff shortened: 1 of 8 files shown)


arsfeld:
Thanks, this is a legitimate guideline point, but I am going to leave it out of this PR and do it separately. Reasoning:

**The markup predates this change.** The `<form>` / `<label>` / `<select>` block is byte-identical at this branch's base (`dc40b43c`), where it arrived in `4860fdb0`, a refactor that split these controls into their own module. This PR touches that block in exactly two ways: it adds `:if={switchable?(@options)}` to the form, and it changes the option comprehension from `SeasonOrder.values()` to `@options`. Neither introduces the pattern being flagged.

**The conversion is not local to this file.** `<.form>` requires `for={@form}`, and the project guideline is that the form is driven by a `to_form/2` assigned in the LiveView rather than built in the template. That means a new assign threaded through four layers: `MediaLive.Show` to `show.html.heex` to `Components.episodes_section` to `SeasonOrderComponents.season_order_controls`, which currently receives only `media_item`, `options`, and `can_update_media`. Separately, `<.input type="select">` renders into `div` plus `label` wrappers with a `fieldset mb-2` container, so the compact inline row here would need its classes retuned to avoid a visual regression.

**There is in-tree precedent for the current form.** Native selects driven by `phx-change` also appear in `lib/mydia_web/live/media_live/index.html.heex` and `lib/mydia_web/live/media_live/show/modals.ex`. This control is an action selector that fires on change and never submits, which is why it was written this way.

Mixing a UI component refactor into a change about which options are offered would make both harder to review, so I would rather land this and convert the control on its own, where a visual diff is the only thing under review.

Issue #525, opened 2026-08-21
Follow-ups from player remote control (#524)


arsfeld:
Follow-ups deliberately left out of #524 (player remote control over p2p). Each was flagged during that work and skipped because it is a real change rather than a tidy-up, not because it is unknown.

The skipped integration test from that PR is not listed here; it is being fixed separately.

### Cast session generation races

- [ ] **`_loadOnRoute`'s success-path `_publish` has no `_connectGeneration` gate.** `player/lib/core/cast/cast_session_manager.dart`, the `_publish(CastSession(...))` immediately after `_store.save(_persisted!)`. A stale `startCast` whose load succeeds can overwrite a newer session. Reproduced during review of #524. Fixing it means threading a generation check into a method three callers share, then deciding what to do with a receiver that loaded successfully but is now unwanted.

- [ ] **`restoreSession` and `_reloadBridgeSession` never participate in `_connectGeneration` at all**, and the latter cancels subscriptions unconditionally. Reachable in practice because `restoreSession` runs unawaited while the UI is already interactive.

### Push and adoption

- [ ] **Both the controller and the target write playback progress during a Push**, which the spec warns against. Falls out of routing `MydiaCastBackend` through the existing backend-agnostic sync path. `stopLocal` pauses the local `Player` but leaves `_progressService` syncing against it. Both writers converge on nearly the same value, so this is wasteful rather than corrupting. See the CodeRabbit thread on `player_screen.dart` in #524.

- [ ] **Adopted sessions show no poster art.** The consumer side is plumbed; `PlayerScreen.describe()` hardcodes `imageUrl: null` (`player/lib/presentation/screens/player/player_screen.dart:3931`) and has no poster to send. Populating it means threading a poster URL through the route and roughly eight navigation call sites. There is a comment at the adoption branch recording this.

### Unbounded event queue

- [ ] **`run_event_dispatcher`'s generic-event channel is unbounded** (`player/rust/mydia_player_p2p/src/lib.rs:701`, `mpsc::unbounded_channel::<Event>()`). A Dart side that never subscribes to `event_stream`, or whose sink has closed, can let the queue grow. This is unbounded by deliberate design since the dispatcher was split, and was untouched by the control-channel fix in #524. Bounding it means choosing a drop-or-coalesce policy for a channel that also carries `Event::Log`, which is a design decision rather than a patch. Exposure is low in practice because `P2pService.initialize()` subscribes unconditionally at startup.

[...]

PR #526, opened 2026-08-21
fix(player): un-skip the remote-control e2e, guard two more disposal races

arsfeld:
Re-enables the remote-control integration test that #524 shipped as `skip: true`, after doing the provider audit that PR asked for.

## Why it was skipped

Nothing in `remote_control_test.dart` was wrong. Three consecutive CI runs each died *before* its body from a different async-disposal race elsewhere in the app. `all_tests.dart` runs every suite in one isolate, and the harness mounts and unmounts a fully-wired `MyApp()` repeatedly, which production never does. That makes it an effective fuzzer for providers that start async work after `build()` without an `ref.mounted` guard.

The reason one escaped exception takes down four unrelated suites is worth stating plainly: an `UnmountedRefException` escaping as an *unhandled* async error during unmount corrupts `TestAsyncUtils`' pump guard for the rest of the process, so every later suite fails on "Guarded function conflict" without running.

## What this fixes

Two providers kick off async work from `build()` via an unawaited `Future.microtask` and were still unguarded:

- `ConnectionDiagnosticsNotifier._loadDiagnostics`
- `UpdateNotifier._initAndCheck`, and `_performCheck` behind it

Both carried a second, subtler shape of the same bug. The `catch` block itself writes `state`, so a disposal that throws out of the `try` lands in the handler and throws *again*, converting a caught error into an unhandled one. Guarding only the success path would have left that live, which is the part worth reviewing.

Also guarded the identical pattern in `recordAttempt`, `recordBatchAttempts`, `clear` and `applyUpdate`. None is reachable from the e2e (the first three have no callers, the last is user-initiated), but they are the same defect and the audit is worth finishing in one pass.

Guard placement and wording follow `connection_provider.dart` and `compatibility_provider.dart`, which already use this convention.

## Verification

`CI / Player E2E` does not run on pull requests, only on push to master and `workflow_dispatch`, so the PR checks below do **not** cover the thing this PR changes. Dispatched by hand against this branch: https://github.com/getmydia/mydia/actions/runs/32494821458

Result is reported in a comment below rather than assumed.

## Known limit

This is an audit, not a structural fix. The suite still shares one isolate, so a future provider written without the guard can reintroduce exactly this failure, and it will again present as an unrelated suite dying. The durable alternative the test file names is giving each suite its own isolate, which costs one app build per file. Not attempted here.

Follow-ups from #524 are tracked separately in #525.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability during connection diagnostics by preventing updates after related screens or services are closed.
  - Improved update checking, installation, and progress handling when interrupted or dismissed.
  - Prevented disposal-related errors during authentication, logout, and peer-to-peer connection updates.
  - Improved reliability when reinitializing connection services.

- **Tests**
  - Re-enabled the remote-control end-to-end integration test.
  - Improved integration test cleanup, including safer handling of pending startup operations and failed assertions.

### DIFF
--- a/player/integration_test/helpers/test_bootstrap.dart
+++ b/player/integration_test/helpers/test_bootstrap.dart
@@ -1,7 +1,11 @@
 import 'package:flutter/widgets.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:graphql_flutter/graphql_flutter.dart';
 import 'package:media_kit/media_kit.dart';
+import 'package:player/app.dart';
+import 'package:player/core/cast/cast_providers.dart';
+import 'package:player/core/graphql/graphql_provider.dart';
 import 'package:player/native/frb_generated.dart';
 
 /// One-time initialization shared by every integration test file.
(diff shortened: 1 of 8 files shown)


arsfeld:
## E2E dispatch results

Three hand-dispatched runs against this branch. Reporting what they said, not assuming.

**Run [32494821458](https://github.com/getmydia/mydia/actions/runs/32494821458)** (provider guards only) — failed, `+1 -1` then a full cascade. It gave the real root cause, which was not an unguarded `ref.mounted` write:

```
Bad state: The provider FutureProvider<GraphQLClient> was disposed during
loading state, yet no value could be emitted
```

raised synchronously from `ElementWithFuture.dispose` inside `StatefulElement.unmount` inside `drawFrame`.

**Run 32497482327** — cancelled. CodeRabbit correctly pointed out the settle helper waited on `hasScheduledFrame`, which reports pending *frames*; a provider blocked on storage or network schedules none, so the loop exited immediately and settled nothing.

**Run [32498448041](https://github.com/getmydia/mydia/actions/runs/32498448041)** (current head) — failed, but the picture changed completely: **`+7 ~2 -2`**, up from `+1`.

- The cascade is gone. All six `p2p_streaming` tests pass.
- `remote_control one player drives another end to end` **executed its body for the first time**, running 2m43s before failing.

### What is still red

Both remaining failures share one cause. `asyncGraphqlClientProvider` restart-loops (29 times in the run, roughly every 6.4s) and never resolves:

```
[asyncGraphqlClientProvider] Starting...
[asyncGraphqlClientProvider] isAuthenticated=true
[asyncGraphqlClientProvider] serverUrl=null
```

It never reaches `Getting graphqlClientProvider...`. The `serverUrl == null` throw sits *after* one more `await`, so the provider stalls rather than settling to an error state.

1. **`remote_control`** fails its own step 1 at `remote_control_test.dart:510`, "Both A and B node IDs should be registered on the server". Player A cannot build a client, so it never registers its node id. This is a genuine assertion failure in a test that had never run before today, not harness damage.
2. **`pairing_flow`** still throws on teardown. `unmountApp` logged `startup providers still loading after 15s`, so the settle worked as designed and correctly gave up: nothing can settle a provider that never resolves.

Pairing does write `serverUrl: 'p2p://mydia'` (`pairing_service.dart:258`), and the degraded-storage overlay in `auth_storage_native.dart` mirrors writes unconditionally and is read first, so a single `Secure storage call failed; using in-memory storage` line does not by itself explain a null read. Why A's `server_url` is absent is the open question.

### Assessment

The structural fix in this PR works and is worth keeping on its own: one escaped error no longer takes out four unrelated suites, which is what made every previous run unreadable. What is left is a real defect the un-skipped test has now surfaced for the first time, and it needs its own investigation rather than another round of teardown hardening.

PR #532, opened 2026-08-22
fix(subtitles): make subtitle downloads work

arsfeld:
Subtitle search returned results but downloading one failed. There were three separate defects behind that single symptom, two in the web UI and one shared with the Flutter player.

The metadata-relay service is not at fault. Verified against the live relay: search, `download-url` and `download` all behave correctly and return valid subtitle bytes.

## A. The web UI crashed on every download

`subtitle_events.ex` called `String.to_integer/1` on the provider's file id. That dates from when OpenSubtitles, whose ids are integers, was the only provider. The two providers enabled by default today produce non-numeric ids: Relay uses base64url strings like `L3N1YnRpdGxlLzM0NjczMzAtODM5MDM4OS56aXA`, Gestdown uses GUIDs. The call raised `ArgumentError` inside `handle_event`, before `start_async`, so the LiveView process died. On a default install that is a 100% failure rate, and OpenSubtitles is off by default because it needs credentials.

The modal now sends a list index and resolves the result server-side from its own assigns, so no provider file id crosses the wire at all and nothing is parsed that can raise. `Integer.parse/1`, not `String.to_integer/1`.

## B. The web UI ignored which provider found the result

The same function passed no provider options, so the downloader always fell back to `provider_type: :relay`. A Gestdown hit, which a default install will produce, was fetched from the relay and rejected. The GraphQL resolver did this correctly; the LiveView never got the same treatment.

Provider resolution now lives in one place, `Subtitles.download_from_result/2`, which both surfaces call. A third caller, the `auto_download` branch of `search_subtitles/2`, was found during review reconstructing the same broken pattern and has been routed through it too. That path is currently unreachable, so it was a latent trap rather than a live bug.

## C. The declared format was a guess, and the downloader trusted it

SubDL and the relay both hardcode `"format" => "srt"` in search results, because the real extension is only visible once the downloaded archive is opened, one request later. Archives legitimately contain `.srt`, `.ass`, `.ssa`, `.vtt` and `.sub`. The downloader validated the downloaded bytes against that declared format, so a real ASS file was rejected outright.

Sampling 60 live relay results across 7 films, 4 of the 57 that downloaded were ASS declared as srt: roughly 7%, failing in both the web UI and the player.

`Format.detect/1` now reads the format from the bytes, and the detected value drives both the stored filename extension and the persisted `format` column, so the delivery layer converts from the correct source. The provider's declared format is a hint and is no longer validated. Detection runs before the temp file is written, so unusable content never reaches disk.

## Notes

- No change to metadata-relay. Detecting from the bytes needs no coordination with a deployed relay, which matters because self-hosted installs upgrade in no particular order.
- No schema change, migration, or backfill. `Subtitle.supported_formats/0` stays `["srt", "ass", "vtt"]` and detection only ever yields those three.
- Non-UTF-8 subtitles are detected correctly and must stay that way: the SRT cue regex deliberately carries no `/u` modifier so PCRE matches byte by byte. A test asserts detection succeeds on content where `String.valid?/1` is false.
- Test fixtures are real bytes captured from the live relay, not synthetic. A fixture in a shape production never produces is exactly how this crash survived 20 passing tests: the old modal test used `file_id: 1`, an integer no enabled provider emits.

## Verification

`mix precommit` passes on the final commit: 8622 tests, 0 failures, Credo and format clean.

## Known follow-up

Three comments still name `generate_subtitle_hash/1`, which became unused and was deleted (`provider/relay.ex:94`, `provider/open_subtitles.ex:271`, `test/.../provider/relay_test.exs:70`). All three only cite it as "same formula as" to document a shared algorithm and none ever called it, so this is stale documentation with no functional coupling.

## Summary by CodeRabbit

* **New Features**
  * Subtitle downloads now detect SRT, VTT, and ASS/SSA formats from file content.
  * Downloaded subtitles use the detected format for filenames and metadata.
  * Downloads route through the originating provider configuration, with fallback support.

* **Bug Fixes**
  * Added clearer errors for unsupported formats, unrecognized content, and invalid selections.
  * Prevented stale or invalid subtitle selections from creating files or records.
  * Improved handling of BOMs, Latin-1 content, and varied line endings.
  * Subtitle downloads now reliably track selections in search results.


arsfeld:
Both findings fixed.

**Anchor ASS detection to header lines** (`80ba09f2c`) — good catch, and the more valuable of the two. `ass?/1` used `String.contains?/2`, so `[Script Info]` anywhere in the 4KB probe won, ahead of the VTT and SRT checks. An SRT whose cue text mentioned it would have been stored with a `.ass` extension and a wrong `format` column, which the delivery layer would then convert from the wrong source. Now anchored with `~r/^\[(?:Script Info|V4\+ Styles|V4 Styles)\]/m`. Deliberately no `u` modifier: this module matches in byte mode so Latin-1 subtitles work, and a test asserts detection succeeds where `String.valid?/1` is false. Covered by a new test using an SRT whose cue *text* contains the literal.

**Reject negative indexes** (`f8e32832c`) — correct. `Integer.parse/1` accepts a leading `-`, so `"-1"` resolved from the end of the list via `Enum.at/2`. Now guarded with `position >= 0`, falling into the same "no longer available" branch. Covered by a new test.

Also noting the earlier red `Test (SQLite)`: the single failure was `FeedbackModalTest` timing out on `render_click`, unrelated to this branch, which touches no feedback code. The same 8622 tests pass locally with 0 failures. This push retriggers a full run.

arsfeld:
Fixed in `93f28a4ce`.

`Integer.parse/1` carries an `is_binary/1` guard, so a `nil`, a JSON number, or a missing key raised `FunctionClauseError` during the call, before the `with`/`else` could catch it, taking the LiveView process down. The clause head now guards on `is_binary(index)` with a fallback clause returning the same "no longer available" flash. Tests added for all three shapes: `nil`, an integer, and an absent key.

Scoped deliberately to this handler. The same exposure exists on other `handle_event` clauses in this module and predates this branch, so widening the diff at merge time did not seem right; that is worth a separate pass.

PR #533, opened 2026-08-22
feat(release): send the changelog's Player notes to TestFlight

arsfeld:
TestFlight's "What to Test" has said `Automated beta build.` to every external beta tester since the iOS lane was written. This replaces it with the release's real notes.

## How it works

`scripts/testflight-notes.sh <version> <sha> <tag>` renders the notes and prints them to stdout, with a `source=bundled|gitlog|placeholder` line on stderr. Three sources, first non-empty wins:

1. The `## Player` section of `priv/changelog/<version>.md`, read from the pinned commit with `git show` rather than the working tree, since a draft release can target an older commit than the ref the workflow runs on.
2. For prereleases, which ship no bundled changelog, commit subjects touching `player/` since the previous tag.
3. A line naming the version. Apple rejects an empty "What to Test" for external distribution, so the output is never empty.

Markdown becomes plain text, because the field renders plain text: subheadings lose their hashes, backticks and bold and links collapse to their content, and `(#524)` references are stripped. Output over Apple's 4000-character cap is trimmed to whole bullets, never mid-sentence, with a link to the release page appended.

`prepare` computes this and publishes it as the `testflight_notes` job output. It is the only job that resolves the version, checks out with `fetch-depth: 0`, and already validates the changelog at the pinned SHA; `player-ios` checks out shallow and pinned. The notes reach the shell through `env:`, never `${{ }}` interpolation into a `run:` body, because they contain backticks and `$`.

The step runs on dry runs too and echoes the result into the job summary, so you can read exactly what testers will see before any platform build finishes.

The Fastfile keeps `Automated beta build.` as a fallback for an unset or unreadable file, so a local `fastlane beta` still works and a failed write degrades to the old behaviour rather than to an upload Apple refuses.

## Verification

`scripts/check-testflight-notes.sh` runs in CI as `Check / TestFlight Notes` and asserts against all 37 bundled changelogs: non-empty, at most 4000 characters, and `source=bundled` for every file that independently greps as containing a `## Player` heading. That last assertion is the load-bearing one. Without it, a broken extraction pattern would fall through to a commit-log summary that is non-empty, under the cap, and wrong.

Current behaviour on real data: 0.13.0 trims from 7334 to 3891 characters with the link tail, 0.13.1 lands at 2194, 0.13.2 at 2628, 0.14.0 at 2054 untrimmed.

Length is counted in characters, not bytes. The notes are not pure ASCII (0.13.0 carries a `·` in a `Watched · Aug 2` badge), and `awk length()` counts characters on gawk but bytes on mawk, which is the default awk on some runner images.

## Two defects caught during review

- The PR-reference stripper was anchored to end of line, so a bullet packing several sentences that each ended in a PR number only had its last one stripped. 0.13.0's notes were shipping `(#344)` and `(#347)` to testers, and the suite could not see it because that assertion was scoped to 0.14.0 only. Now a global strip, with an assertion covering 0.13.0.
- `previous_tag()` aborted the whole script under `set -e` when the filtered tag list came back empty, because `pipefail` propagates `grep`'s exit 1 even though `head` succeeds. It never reached the placeholder fallback. Unreachable in this repository, which has 70 tags, but latent.

## Not included

The Play Store `internal` lane still passes `skip_upload_metadata: true`, so Android release notes remain blank for the same underlying reason. The `prepare` output added here makes that a small follow-up.

## Summary by CodeRabbit

* **New Features**
  * TestFlight releases now receive automatically generated notes from player changelog updates.
  * Notes are plain text, capped at Apple’s 4,000-character limit, and may include a release link.
  * Prereleases can fall back to player commit summaries when changelog details are unavailable.
  * Release preparation summaries display generated notes, including during dry runs.
  * Added a fallback message when notes cannot be generated or read.

* **Tests**
  * Added automated validation for note generation, formatting, truncation, and fallback behavior in continuous integration.

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -93,6 +93,25 @@ jobs:
       - name: Assert every Apple deployment target agrees
         run: ./scripts/check-apple-deployment-targets.sh
 
+  testflight-notes:
+    name: Check / TestFlight Notes
+    runs-on: ubuntu-latest
+
+    steps:
+      # fetch-depth: 0 because the commit-log fallback resolves a previous tag
+      # and walks a commit range. A shallow checkout has neither.
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # The notes this produces become TestFlight's "What to Test", which Apple
+      # caps at 4000 characters and rejects when empty. Both properties are
+      # invisible until a release is dispatched, so they are asserted here
+      # against every bundled changelog instead. Needs only the checkout.
+      - name: Assert the store release notes hold their contract
+        run: ./scripts/check-testflight-notes.sh
+
   # Both NixOS VM tests run on push only. They are the slowest and least
   # reliable jobs in CI, combining a QEMU boot with a full Nix build on top of
   # the FlakeHub cache 503s that can hit any Nix job, and breaking the NixOS
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -640,12 +640,23 @@ jobs:
           bundler-cache: true
           working-directory: ${{ env.PLAYER_PATH }}/ios
 
+      # The notes reach the shell through env, never through ${{ }} substitution
+      # into this body: changelog text contains backticks and $, which would be
+      # both an injection risk and a corruption bug.
+      - name: Write TestFlight notes
+        if: ${{ !inputs.dry_run && env.APP_STORE_CONNECT_API_KEY_ID != '' }}
+        env:
+          APP_STORE_CONNECT_API_KEY_ID: ${{ secrets.APP_STORE_CONNECT_API_KEY_ID }}
+          NOTES: ${{ needs.prepare.outputs.testflight_notes }}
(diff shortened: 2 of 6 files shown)


arsfeld:
@coderabbitai full review

arsfeld:
Both findings were real and are fixed in e3efd4c05.

**Commit-log notes not normalized.** Confirmed against real data before fixing: squash merges always append the PR number, so `Address PR review feedback (#475)` is an ordinary subject in this range and testers would have seen the reference. Piped that source through `to_plain_text`.

Worth noting on the test: the obvious assertion would have been vacuous. Reusing the existing `v0.13.0..HEAD` range does not work, because it runs to 120+ subjects and the marked-up one sits at line 120 while the 4000-character budget only keeps 61 lines, so it never reaches the output. The check now pins `v0.13.2..7f080ff3`, which is 46 subjects with the marked-up one first. Verified it fails without the fix and passes with it.

**Template injection in the prepare step.** Agreed, and it was inconsistent of me: the notes text was already passed through `env:` for exactly this reason while version/sha/tag were not. The tag is a dispatch input, so one containing `$(...)` would have been command substitution. All three go through `env:` now and the `run:` body has no `${{ }}` left.

arsfeld:
@coderabbitai review

PR #556, opened 2026-08-25
Request posters and detail popup

arsfeld:
## Summary

When managing media requests, the cards showed only a title, so an admin had nothing to judge a request by and no way to see more without leaving the page. This adds a poster to every request card and makes the title open the same rich detail popup Discovery and the Dashboard already use.

Both `/admin/requests` and `/requests` are covered. They previously duplicated about ninety lines of card markup and three formatting helpers apiece; both now render through one shared component.

## What changed

- **`media_requests.poster_path`** (new nullable `:text` column). Populated at request creation by both existing creation paths, which already had the value and were discarding it. Rows created before the column existed fill in lazily on first view, so no data migration reaches the network at boot.
- **`MydiaWeb.MediaRequestComponents.request_card/1`**, a shared card with `:badges`, `:details` and `:actions` slots. Each page keeps its own differences in slot content rather than the component growing a flag per difference.
- **Metadata resolution routes on whichever id the row stored.** A TV request made while the instance is set to TVDB metadata carries only a `tvdb_id`, so a TMDB-only lookup would have left every TV request on those instances with no poster and no popup.
- **An optional `:actions` slot on `TrendingDetailModal`.** Its default footer renders unchanged when the slot is absent, so the Dashboard and Discovery callers are untouched. The requests pages supply Approve, Reject, View in Library and Close instead, because "Add to Library" there would create the media item while leaving the request stuck in `pending`.
- **Approve and Reject from inside the popup** close the detail dialog in the same render that opens the approve dialog, so two dialogs are never on screen at once. That matters because the modal's `phx-window-keydown` Escape binding is not scoped by visual stacking.

## Notes for review

- A request with only an `imdb_id` is permitted by the schema. Those render the placeholder poster and a non-clickable title; one predicate gates the click target, the backfill and the popup so they cannot disagree.
- The poster backfill marks request ids attempted at send time and runs at most once per row per connected socket. Marking on completion instead would loop forever on any row whose fetch permanently fails, since `poster_path` stays nil and the row keeps qualifying.
- Relay failures degrade to the existing placeholder rather than retrying in a loop; the next page visit retries.

## Testing

- `mix test` on SQLite: 13 doctests, 9089 tests, 0 failures
- `DATABASE_TYPE=postgres mix test` across the touched suites: 76 tests, 0 failures
- `mix format --check-formatted`, `mix compile --warnings-as-errors`, `mix credo --strict`: all clean

## Summary by CodeRabbit

* **New Features**
  * Added request detail popups with metadata, actions, and library navigation.
  * Added poster artwork to media requests, including automatic backfilling and placeholders.
  * Improved request cards with status badges, dates, media details, and context-aware actions.
  * Added library links for approved requests with available media.
* **Bug Fixes**
  * Prevented repeated metadata and poster lookups for unresolved requests.
  * Improved handling of requests lacking provider identifiers.
  * Prevented individual metadata or poster failures from interrupting request displays.

### DIFF
--- a/lib/mydia/media/media_request.ex
+++ b/lib/mydia/media/media_request.ex
@@ -17,6 +17,7 @@ defmodule Mydia.Media.MediaRequest do
           tmdb_id: integer() | nil,
           tvdb_id: integer() | nil,
           imdb_id: String.t() | nil,
+          poster_path: String.t() | nil,
           status: String.t(),
           requester_notes: String.t() | nil,
           admin_notes: String.t() | nil,
@@ -41,6 +42,7 @@ defmodule Mydia.Media.MediaRequest do
     field :tmdb_id, :integer
     field :tvdb_id, :integer
     field :imdb_id, :string
+    field :poster_path, :string
     field :status, :string, default: "pending"
     field :requester_notes, :string
     field :admin_notes, :string
@@ -68,6 +70,7 @@ defmodule Mydia.Media.MediaRequest do
       :tmdb_id,
       :tvdb_id,
       :imdb_id,
+      :poster_path,
       :requester_notes,
       :requester_id
     ])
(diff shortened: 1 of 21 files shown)


arsfeld:
Thanks CodeRabbit. Addressed two of the four findings and declining two, with reasoning.

**Fixed: provider work blocking the LiveView process** (b4947d4d6)

Correct, and it matches an established pattern here rather than being a style preference: `lib/mydia_web/live/discover_live/index.ex:389-399` already documents this exact reasoning for `start_async`, and `start_async` is used in four other places. This branch had modeled its handler on Discovery's *other*, still-inline one. Both `handle_info` clauses in both request LiveViews are now `start_async/3` with matching `handle_async/3` callbacks, and the misleading comments claiming that sending to `self()` avoids freezing the process have been corrected: deferring to `handle_info` only lets the first render paint, it does not keep the process responsive afterwards.

**Resolved by that same change: converting provider exceptions to error tuples**

That finding was written against the previous head, where a raise would have propagated out of `handle_info` and taken down the LiveView, so the premise was right at the time. Under `start_async` a raise surfaces as `{:exit, reason}` to `handle_async/3`, which logs it and clears `detail_loading` rather than crashing the process. `backfill_one/1` also keeps its own per-row `try/rescue`, following the `Mydia.Metadata.Provider.Relay.fetch_all_season_episodes/3` precedent.

**Declining: extract the duplicated LiveView glue into a shared module**

The duplicated code is socket orchestration (mount assigns, event and async callbacks), not query logic. The queries themselves already live in `Mydia.MediaRequests`, which is what the "never duplicate query logic across LiveViews" guideline targets. At two occurrences this also sits under the project's own stated bar of three or more before extracting, and the two pages' `load_requests/1` genuinely differ (My Requests filters by `requester_id`). An independent review of the whole branch reached the same conclusion, with the note that if a third request-like page is ever added, this is the first thing that should be pulled out rather than copied again.

**Declining: move metadata resolution and the backfill into a context module**

`fetch_request_metadata/1`'s TMDB branch delegates to `MediaAddHelpers.fetch_detail_metadata/2`, which is itself a web helper, so moving the resolver into `Mydia.MediaRequests` would point a context module at `MydiaWeb`. The suggestion acknowledges this and proposes extracting the shared TMDB resolver out of `MediaAddHelpers` first, but that touches the Dashboard and Discovery call sites and is well outside this PR. The persistence half already is in the context: `MediaRequests.update_poster_path/2`. Worth doing as its own change.

PR #596, opened 2026-08-28
Classify movie extras instead of presenting them as versions

arsfeld:
## Problem

Bonus features sitting loose in a movie folder were imported as ordinary media files, so the movie page presented them as versions of the feature. On galactica: Ratatouille has 37 files, of which 36 are extras sitting beside one real 3.3 GB BluRay rip. Fantastic Mr. Fox has 24, Mulan 22, Monsters University 21.

`SampleDetector` documents three detection layers and all three missed these files. The folder layer needs a Plex extras subdirectory; these sit flat in `Ratatouille (2007)/`. The filename layer is `$`-anchored on suffixes like `-featurette`, so `Chez Gusteau_new.mkv` and `Your Friend the Rat - Short Film.mkv` never matched. The third layer, `sample_by_duration?/2`, is the only one that would catch a 3 MB 480p deleted scene, and it had **zero production call sites** — exercised only by its own doctests and one unit test.

That third layer was dead for a structural reason rather than an oversight: duration is not available where detection ran. The scanner filters at scan time; `metadata.duration` is written later by `FileAnalysis`. At the scan gate the row does not exist yet.

## What this does

Two-phase classification, with three new nullable columns on `media_files` (`extra_kind`, `extra_source`, `extra_checked_at`).

**Phase one, scan time.** The scanner previously dropped detected extras silently. It now persists them with a classification, so a file Mydia decided to ignore leaves a record instead of being re-detected and re-dropped on every subsequent scan.

**Phase two, after analysis.** A new recurring worker (`Mydia.Jobs.ExtraClassification`) batches by movie and demotes files running at under half their movie's published runtime. This is the production call site `sample_by_duration?/2` never had. Existing libraries need no backfill: every row is `extra_checked_at IS NULL` after the migration, which is the worker's own selection criterion.

**One query seam.** `where: is_nil(trashed_at)` was hand-copied across many modules. Two shared queries now exist — `MediaFile.active/0` (extras included, for the movie page) and `MediaFile.versions/0` (extras excluded) — and every "which file belongs to this item" caller routes through the latter.

**UI.** Extras collapse into a disclosure under the file list showing each file's kind and how it was detected, with controls to promote an extra to a version or demote a version to an extra. Operator decisions are sticky: the worker never reconsiders a row whose `extra_source` is `:operator`.

## Calibration

The 0.5 ratio is measured, not guessed. Over 354 movie files on galactica the distribution is bimodal with an empty band between 0.30 and 0.50: 144 files below 0.30, one at 0.499, then nothing until 0.614. The nearest genuine feature above the line is *LEGO Frozen: Operation Puffins* at 0.896 — a 16.1 minute film with an 18 minute published runtime — so the threshold has roughly 0.4 of headroom to the closest real film. Everything from 0.90 to 0.97 is a legitimate TS/CAM/PreDVD rip whose published runtime runs slightly long.

The rule is one-sided by design: a file longer than the reference is never demoted. *Spider-Man: Into the Spider-Verse* carries an Alternate Cut at 1.227 and four *Land Before Time* files sit between 1.05 and 1.13; a symmetric tolerance would have demoted all of them.

A second constant, `@rescue_floor` 0.30, bounds the last-version-survives invariant so it rescues only a file that is plausibly a feature. Without it a folder holding only a three minute clip would promote that clip to the movie's version and the movie would falsely report as owned.

## Measured effect

145 of 354 movie files reclassified. Movies presented as multi-version drop from 39 to 16, with 160 landing on exactly one version. The invariant fires exactly once, on `Enchanted`, whose only file is a different film misfiled into the folder.

Three extras stay versions, all on one movie: Fantastic Mr. Fox's `Roald Dahl Reads Fantastic Mr. Fox` (0.614), `Fantastic Mr. Dahl` (0.706) and `Animatic version of the film` (0.864). Catching the animatic would require a threshold above LEGO Frozen's 0.896, which would demote real films — the manual demote control covers these instead.

## Regressions this closes

Persisting extras made several "does this item have files" queries start counting bonus content as the item. Each is fixed here, and none of these states was reachable before:

- `movies_needing_search/1` and `load_monitored_movies_without_files/0` left-join files and keep only items with `count == 0`, so an extras-only movie was excluded from search **permanently**.
- `Downloads.Queue` skipped the download entirely, logging "media files already exist for movie".
- The same pattern on the TV side: the scanner's path layer is media-type agnostic, so a persisted `S01E01-trailer.mkv` made episodes look owned.
- `episode_has_media_file?/1` gated season-pack imports, so a persisted episode trailer made the importer skip the episode's real file while the (fixed) search path kept re-grabbing it. Downloaded repeatedly, imported never.
- The library page's Downloaded badge, the on-deck rail, and the upgrade sweep.

Analysis workers deliberately still see extras — analysis is what produces the duration the classifier reads. `import_groups`, `prune/grouping`, trash cleanup and the movie page's own loader are likewise untouched by design.

## Upgrade note

The first scan after this ships is not silent. Folder-detected extras were never persisted, so `detect_changes/3` re-flagged and re-dropped them every scan; now one scan creates rows for all of them at once. On a large library that is a single batch of new rows.

## Known follow-up

`import_run.ex`'s `sample_or_extra?/1` is a second scan-and-create pipeline behind the Import Media flow that still drops folder- and filename-matched extras rather than flagging them. It is unchanged pre-existing behaviour rather than a regression, the spec scopes the import path out, and the duration-detected case this PR targets is unaffected. Left for a follow-up rather than newly persisting extras on a path whose queries have not been audited.

## Verification

Full suite green on both adapters: SQLite 9456 tests / 0 failures, PostgreSQL 9453 tests / 0 failures. `credo --strict`, `format --check-formatted` and `compile --warnings-as-errors` all clean. TV is out of scope for classification; the worker is gated to `media_items.type == "movie"`.

## Summary by CodeRabbit

* **New Features**
  * Automatically identifies bonus content such as trailers, samples, and deleted scenes.
  * Media details now separate versions from extras, with options to promote or demote files.
  * Extra classifications are periodically updated in the background.

* **Bug Fixes**
  * Bonus content no longer counts as the main movie or episode file.
  * Playback, thumbnails, subtitles, upgrades, and watch status now focus on primary versions.
  * Movies and episodes containing only extras remain eligible for missing-file searches.
  * Scanned bonus content is retained and labeled instead of being discarded.


arsfeld:
Thanks — both findings reviewed. One fixed, one declined with reasoning.

**Authorization bypass in `reclassify/3` — fixed** in `f6138d81`.

Valid, and worth being specific about why: this module already guards its destructive handler with `Authorization.authorize_delete_media/1` at `file_events.ex:339`, so the new handlers were inconsistent with the file's own convention, not just with a general standard. `reclassify/3` now requires `authorize_update_media/1` and requires the id to belong to the media item on screen, both checked before anything is fetched or written.

One implementation note for anyone reading the diff: the obvious scoping check — comparing `media_item_id` — would be wrong here. A TV file carries `episode_id` with a NULL `media_item_id` and reaches its show only through the episode, so that check would silently refuse every episode file. The fix scopes against `all_media_files/1`, which the page already uses and which merges both.

Two tests added: an unauthorized user cannot promote or demote, and a file id belonging to a different media item is refused. The second is the one that matters, since it does not depend on role configuration.

Not changed: `mark_file_preferred/2` at line 463 has the identical gap. It is pre-existing and outside this PR's scope, so it is left alone rather than fixed opportunistically here.

---

**Excluding extras from `/stream/:id` — declining, deliberately.**

The extras filter in this PR is applied to queries that *select* a file on an item's behalf — "which file is this movie", "does this episode have files". `stream/2`'s direct-id clause is not selection; the caller has named one specific file, and an explicit id is explicit intent.

That is the same line this PR already draws elsewhere: `ThumbnailGeneration`'s `library` and `missing` modes are filtered because they choose rows themselves, while `single` and `batch` are deliberately left unfiltered because they take explicit ids. Filtering the direct stream route would make this PR inconsistent with its own rule.

It is also unreachable in practice today. Extras are excluded from every path that hands a file id to a client, so a caller only reaches this route with an extra's id if it constructed one deliberately — and if a future change adds extras playback, this is the route it would use.

Happy to revisit if you can point at a path that surfaces an extra's id to a client where it shouldn't; that would make it reachable and change the answer.

PR #599, opened 2026-08-29
fix(config): stop dropping database-backed settings at every boot

arsfeld:
Configuration loaded in `Mydia.Application.start/2` runs before `Mydia.Repo` is
started, and `load_database_config/0` rescued the resulting error into an empty
layer. Every `ConfigSetting` row was silently dropped on every boot, and nothing
re-merged it afterwards, so a setting saved in the admin UI worked until the
next restart and then reverted with nothing logged.

FlareSolverr is where this surfaced: its admin row read the rows directly and
showed Enabled while its client read the cached config and answered
"FlareSolverr is disabled. Enable it in configuration."

## Changes

- `Loader.load/1` takes a `:sources` option; `Application.start/2` uses
  `[:yaml, :env]` and no longer needs a rescue at all.
- New `Mydia.Config.Bootstrap` child, after `Ecto.Migrator`, re-merges all four
  layers. Placement after the migrator is required: `config_settings` does not
  exist on a fresh install until it has run. It is gated `skip:` in test, where
  the supervisor process owns no SQL Sandbox connection.
- `load_database_config/1` no longer rescues the bare `RuntimeError` Ecto raises
  for an unstarted Repo, and now covers `Postgrex.Error` alongside
  `Exqlite.Error` (a pre-existing gap on a first-class adapter).
- The FlareSolverr admin row reads values from the merged config only. Rows are
  consulted to label ENV/DB/Default and nothing else, so the badge and the
  client can no longer disagree.

## Verification

Beyond the unit tests, this was confirmed end to end against a real
dev-environment boot: with `flaresolverr.enabled=true` and a URL in
`config_settings`, the boot-time merged config now reports `enabled: true` with
the URL set, and `health_check/0` returns a connection error for the fake probe
host rather than `{:error, :disabled}`. Before the fix the same probe produced
`enabled: false`, `url: nil`, `{:error, :disabled}`.

Full suite: 9561 tests, 0 failures. Credo `--strict` clean, format clean.
The Postgres suite was not run locally (the devenv Postgres service was not
started); CI covers both adapters.

Regressed in a51171e60, which moved `FlareSolverr.get_config/0` off
`Application.get_env(:mydia, :flaresolverr)`. That key is set unconditionally in
`config/runtime.exs`, which is why 1.3 worked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Runtime configuration now incorporates database settings after migrations complete during startup.
  * Admin indexer settings consistently reflect the active runtime configuration.
  * Improved handling of unavailable or uninitialized databases without disrupting startup.

* **Improvements**
  * Configuration loading can selectively use YAML, database, and environment sources.
  * Invalid database settings no longer replace the last valid runtime configuration.
  * FlareSolverr settings now consistently reflect the active runtime configuration across the application.

### DIFF
--- a/lib/mydia/application.ex
+++ b/lib/mydia/application.ex
@@ -316,7 +316,11 @@ defmodule Mydia.Application do
   end
 
   defp load_config! do
-    Mydia.Config.Loader.load!()
+    # Phase one: no database layer. Mydia.Repo is a child of the tree this
+    # config is used to build, and on a fresh install config_settings has not
+    # been migrated. Mydia.Config.Bootstrap merges the database layer after
+    # Ecto.Migrator.
+    Mydia.Config.Loader.load!(sources: [:yaml, :env])
   end
 
   defp ensure_default_quality_profiles do
--- a/lib/mydia/application.ex
+++ b/lib/mydia/application.ex
@@ -99,6 +99,11 @@ defmodule Mydia.Application do
       {Mydia.Release.MigrationBackup, skip: skip_migrations?()},
       {Ecto.Migrator,
        repos: Application.fetch_env!(:mydia, :ecto_repos), skip: skip_migrations?()},
+      # Re-merges the database configuration layer, which start/2 could not read
+      # because it runs before this tree exists. Must sit after the migrator: on
+      # a fresh install config_settings does not exist until the migrator has
+      # run. Returns :ignore once done. See the module doc.
+      {Mydia.Config.Bootstrap, skip: skip_config_merge?()},
       # Releases import runs whose coordinator died with the previous node.
       # Must run after the migrator (it queries import_runs) and before the
       # Oban child below, because "no queue has started yet" is exactly what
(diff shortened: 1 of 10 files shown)


arsfeld:
Follow-up identified during review, not blocking this PR but worth tracking.

**A single malformed `config_settings` row invalidates the entire merge.**

`Loader.validate/1` runs one whole-struct `Schema.changeset/2` over the merged
map, so one bad row rejects the whole database layer rather than just the
offending field. `Config.Bootstrap` handles that safely (logs at `:error`, keeps
the phase-one config, never crashes), but the user-visible effect is that every
database-backed setting drops on the next boot, which looks exactly like the bug
this PR fixes.

This was unreachable before, because the database layer was never merged at boot
at all. Making the merge work is what makes it reachable.

Concrete trigger: `flaresolverr.enabled=true` with no URL, which
`schema.ex:604-611` rejects. The admin UI's own changeset blocks creating that
state today, so it would take an older version or a write that bypassed the UI.
An instance that accumulated such a row while the ordering bug was masking the
database layer would hit this on upgrade.

Suggested fix: per-field skip-and-log instead of whole-layer rejection.

Separately: no CI job boots the real supervision tree with `Config.Bootstrap`
live on a PR. `mix test` always sets `start_health_monitors: false` and skips it,
and `ci-player-e2e.yml` (which does boot the release) runs on push to master
only. That is a pre-existing gap shared by `MigrationBackup` and the health
children, not something this PR introduces. The pre-merge boot evidence here is a
manual `./dev mix run` against a real dev boot: with the setting in the database,
the merged config reported `enabled: true` and `health_check/0` returned a
connection error for a deliberately fake host rather than `{:error, :disabled}`.

PR #616, opened 2026-08-29
fix(media): make the Media Files card usable on mobile

arsfeld:
The Media Files card on a movie page rendered each file's full absolute path in `font-mono break-all` inside a flex row next to a fixed ~200px strip of six icon buttons, at every viewport width. On a 375px phone that left the path roughly 90px of column, where `break-all` shredded it into a tall stack of single characters beside 32px buttons too small to hit.

Rows now show the basename with `truncate`, carrying the full path on a `title` attribute. The full path is still rendered in the file details modal that every row already links to. Below the `sm:` breakpoint the button strip drops to its own full-width line and the icon buttons grow from 32px to daisyUI's default 40px.

Also here: the metadata row's `gap-4` becomes `gap-x-3`, codec names run through the existing `shorten_codec/1` helper so "Dolby Digital Plus" renders as "DD+", the extras list gets the same treatment, and the details modal falls back to "Unknown file" instead of rendering a blank path box for an orphaned row.

## A regression that only measurement caught

The first five commits introduced a worse bug than the one being fixed, and it took driving a real browser to see it. `truncate` sets `white-space: nowrap`, which only clips against a width-constrained ancestor. daisyUI's `.menu` ships `width: fit-content; flex-flow: column wrap`, so an unbreakable 82-character basename inflated the row to 712px against a 351px card and gave the whole page a horizontal scrollbar on mobile. The original `break-all` was ugly, but it wrapped, so it never overflowed.

Every other non-popover `.menu` list in this codebase already carries `w-full`; these two lists were the anomaly. The last commit adds that, plus `flex-nowrap` and `items-stretch` to defeat the inherited wrap and a daisyUI selector forcing `align-items: center`.

## Measured in headless Chromium

375x812, versions row, on a file with an 82-character basename:

| | before | after five commits | shipped |
|---|---|---|---|
| label height (line height 22.75) | 386.75 (~17 lines) | 22.75 (1 line) | 22.75 (1 line) |
| label actually clips | no, wraps | **no, inflates** | **yes** |
| row width vs card clientWidth 351 | 319 | **712.5** | 319 |
| document scrolls horizontally | no | **yes** | no |
| icon buttons | 32x32 | 40x40 | 40x40 |

Extras row: 569.7 to 319 against the same 351px card.

1440x900 desktop: row still side-by-side, buttons back to 32x32, row 728 vs card 776, no scroll.

The probe was falsified before being trusted: run against the pre-change markup it throws its "target missing" error, and an adapted probe reproduces the old 17-line, 32px, no-scroll numbers.

## Tests

31 component tests across `media_files_section_test.exs` and `modals_test.exs`; 462 tests green across `test/mydia_web/live/media_live/`. The new tests query the label element and read its text and `title` separately, because a whole-document substring assert is satisfied by the `title` attribute alone and would pass even when the visible text is wrong.

`episode_file_row/1` is deliberately untouched: it already renders a truncated basename and degrades correctly.

## Summary by CodeRabbit

- **Improvements**
  - Media file lists now fit within their cards without expanding for long filenames.
  - Filenames display compactly, with the full path available on hover.
  - Codec information is shortened for readability, with “Unknown” shown when unavailable.
  - File action buttons and technical details have improved responsive spacing and alignment.
  - File details show the full path when available and clearly identify orphaned files.

- **Tests**
  - Added coverage for filename display, paths, codec labels, extra files, and orphaned media files.

### DIFF
--- a/lib/mydia_web/live/media_live/show/components.ex
+++ b/lib/mydia_web/live/media_live/show/components.ex
@@ -879,11 +879,14 @@ defmodule MydiaWeb.MediaLive.Show.Components do
             </summary>
             <ul class="menu bg-base-100 rounded-box p-0 mt-2">
               <li :for={file <- @extras} id={"extra-#{file.id}"}>
-                <div class="flex items-center justify-between gap-4 p-4 rounded-none">
-                  <div class="flex-1 min-w-0 flex flex-col gap-1">
-                    <p class="text-sm font-mono break-all">
-                      {Mydia.Library.MediaFile.display_path(file) ||
-                        Mydia.Library.MediaFile.display_name(file)}
+                <div class="flex flex-col gap-2 p-4 rounded-none sm:flex-row sm:items-center sm:justify-between sm:gap-4">
+                  <div class="min-w-0 sm:flex-1 flex flex-col gap-1">
+                    <p
+                      id={"extra-name-#{file.id}"}
+                      class="text-sm font-mono truncate"
+                      title={Mydia.Library.MediaFile.display_path(file)}
+                    >
+                      {Mydia.Library.MediaFile.display_name(file)}
                     </p>
                     <div class="flex flex-wrap gap-2 text-xs text-base-content/60">
                       <span class="badge badge-ghost badge-sm">{file.extra_kind}</span>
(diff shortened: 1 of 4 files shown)


arsfeld:
CodeRabbit's one finding is applied in d1631cfce. Measured at 375x812 the extras promote button was 89.3x32 beside 40x40 version-row actions; it now takes the default size below `sm` and `btn-sm` from `sm` up, matching them.

PR #643, opened 2026-09-01
ci(e2e): run the player E2E on every PR

arsfeld:
Re-enables `CI / Player E2E` on pull requests, after first fixing the reason it was expensive.

## What the investigation found

The workflow was moved to push-only on 2026-08-12 because every backend PR paid a from-source server build of roughly 20 minutes. Two things turned up that the original decision could not have known.

**Every BuildKit cache flag in this job has always been a silent no-op.** The `--cache-from` / `--cache-to type=gha` flags live in raw `docker buildx build` calls inside `run:` steps, and GitHub injects `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_CACHE_URL` into actions only, never into `run:` steps. BuildKit skips the cache backend without failing, so nothing ever pointed at the problem.

Confirmed on run 33444861643, a green master push: the only build that logged `importing cache manifest` was `Publish / Toolbox`, the one job using `docker/build-push-action`. The server build imported no manifest and produced zero `CACHED` layers across 12m30s. The `player-toolbox` cache scope had exactly one entry, ever.

So the build everyone was avoiding was a fully cold build. Its layer breakdown:

| Layer | Time |
| --- | --- |
| `mix compile` (699 `.ex` files plus `mydia_p2p` and `mydia_subsync` in release mode, then two WASM plugins) | 7m51s |
| flutter-builder: `pub get` + `build_runner` + `flutter build web --release` | 3m07s |
| `apk add` builder deps | 2m05s |
| `mix deps.compile` | 1m59s |

**Master passed 2 of its last 12 runs.** Three distinct causes: a flutter_rust_bridge codegen 2.12/2.13 mismatch (5 failures, deterministic, fixed by #636); `e2e_relay` dying at boot with `sys_sigaltstack(): Failed to set alternate signal stack`, exit 134, before any test ran (3 failures, hit a docs-only PR and a CI-config PR, so environmental); and `p2p_streaming_test.dart` instability (3 failures).

## Changes

1. **Expose the Actions runtime** so `type=gha` works from `run:` steps. This is the fix that makes every other cache flag in the file mean something.
2. **`publish-toolbox` no longer waits on the test job.** It builds a Flutter/GTK/mpv toolchain the E2E tests never validate, since the test container bind-mounts the repo and builds the player at runtime. Gating it froze both the registry tag and its cache scope whenever tests went red, leaving the next push in the same cold state that caused the failure.
3. **Retry the two flakes at the right granularity.** Dependency startup gets 3 attempts, because a relay crash costs seconds. The test run gets one retry from a fully torn-down stack, because `mydia` is seeded at boot and the tests mutate it. A pass on the retry logs a `::warning` marking the run FLAKY, so a retried pass never reads as clean.
4. **Restore the `pull_request:` trigger.** Everything downstream was already written for PRs and simply unreachable.

A follow-up commit corrects comments this branch made stale, and narrows the relay cache write to `push` so a manual `workflow_dispatch` from a feature branch cannot seed the shared master scope.

## Expected timings

| PR touches | Path | Job time |
| --- | --- | --- |
| player, docs, CI only | pull all three images | ~10m |
| backend paths | build server, pull the rest | ~19m (est.) |
| nothing under `relevant` | skipped by the filter | seconds |

## Still unproven

The retry logic cannot be exercised without a real failure. It ships untested against a live flake and gets its first real test the next time `sigaltstack` fires or `p2p_streaming_test.dart` flakes. Watch for the `::warning` marking a run FLAKY rather than reading a green check as clean.

Measurements to follow in a comment. The `~19m` figure above is an estimate derived from cold layer timings; the forced-build run will replace it with a real number.

## Summary by CodeRabbit

- **Bug Fixes**
  - Pull requests are now automatically covered by end-to-end validation.
  - Improved build caching helps make test and toolbox workflows more reliable.

- **Reliability**
  - Dependency startup and failed end-to-end tests now retry automatically.
  - Flaky test activity is clearly flagged, with retry logs preserved for troubleshooting.

- **Workflow Improvements**
  - Toolbox publishing can proceed independently of end-to-end test completion.
  - Manual and push-triggered workflow options remain available.

### DIFF
--- a/.github/workflows/ci-player-e2e.yml
+++ b/.github/workflows/ci-player-e2e.yml
@@ -277,19 +277,19 @@ jobs:
           # docker-sqlite scope: the GHA cache is capped at 10 GB with LRU
           # eviction, and docker-sqlite is populated by a different workflow,
           # so writing to it here would evict entries that workflow depends
-          # on. player-toolbox and relay-master are different: only jobs in
-          # this file ever write to them, so there is no cross-workflow
-          # contention. Populating them on non-PR builds means even a
-          # timed-out first run (see timeout-minutes above) leaves a warm
-          # cache behind, so the next push is fast instead of repeating the
-          # same cold build. Restricted to non-PR events to match the two
-          # registry tags themselves, which publish-toolbox and ci-relay.yml
-          # only ever push on push events; a PR build's layers are not
-          # something master's cache should be seeded from.
-          TOOLBOX_CACHE_TO=()
+          # on. relay-master is an exception: only ci-relay.yml and this job
+          # ever write to it, and ci-relay.yml runs only on metadata-relay
+          # paths so concurrency is rare. publish-toolbox owns the
+          # player-toolbox scope exclusively since Task 2 broke the deadlock
+          # that justified populating it from here; this job reads the scope
+          # but does not write it. Populating relay on non-PR builds means
+          # even a timed-out first run leaves a warm cache behind, so the
+          # next push is fast instead of repeating the same cold build.
+          # Restricted to non-PR events to match the registry tags themselves,
+          # which ci-relay.yml only ever pushes on push events; a PR build's
+          # layers are not something master's cache should be seeded from.
           RELAY_CACHE_TO=()
           if [ "${{ github.event_name }}" != "pull_request" ]; then
-            TOOLBOX_CACHE_TO=(--cache-to type=gha,mode=max,scope=player-toolbox)
             RELAY_CACHE_TO=(--cache-to type=gha,mode=max,scope=relay-master)
           fi
 
@@ -310,7 +310,6 @@ jobs:
             docker buildx build \
               --file player/Dockerfile.test \
               --cache-from type=gha,scope=player-toolbox \
-              "${TOOLBOX_CACHE_TO[@]}" \
               --load --tag "$TOOLBOX" .
           fi
 
--- a/.github/workflows/ci-player-e2e.yml
+++ b/.github/workflows/ci-player-e2e.yml
@@ -214,7 +214,7 @@ jobs:
       # environment of later `run:` steps.
       - name: Expose GitHub Actions runtime for BuildKit
         if: steps.should.outputs.run == 'true'
-        uses: crazy-max/ghaction-github-runtime@v3
+        uses: crazy-max/ghaction-github-runtime@3cb05d89e1f492524af3d41a1c98c83bc3025124 # v3
 
       # The resolve steps below replace the former `docker compose build`.
       # They pull by default and build only what a PR actually changed, which
@@ -403,6 +403,10 @@ jobs:
               if compose up -d --wait relay mydia; then
                 echo "::endgroup::"
                 echo "::notice::Dependencies healthy on attempt ${attempt}/3."
+                if [ "$attempt" -gt 1 ]; then
+                  echo "::warning::Dependencies needed ${attempt} attempts to start. This run is FLAKY, not clean. See e2e-deps-attempt-*-logs.txt in the artifacts."
+                  echo "E2E_RETRIED=true" >> "$GITHUB_ENV"
+                fi
                 return 0
               fi
               echo "::endgroup::"
@@ -427,6 +431,7 @@ jobs:
             if [ "$status" -eq 0 ]; then
               if [ "$cycle" -gt 1 ]; then
                 echo "::warning::Player E2E passed only on cycle ${cycle}. This run is FLAKY, not clean. See e2e-cycle-1-logs.txt in the artifacts."
+                echo "E2E_RETRIED=true" >> "$GITHUB_ENV"
               fi
               exit 0
             fi
@@ -448,8 +453,14 @@ jobs:
           docker compose -f compose.player-e2e.yml logs relay > relay-logs.txt 2>&1 || true
           docker compose -f compose.player-e2e.yml logs test > test-logs.txt 2>&1 || true
 
-      - name: Upload logs on failure
-        if: failure() && steps.should.outputs.run == 'true'
+      - name: Upload logs on failure or retry
+        # Not purely failure()-gated: a retried success (E2E_RETRIED=true,
+        # set by the "Run player E2E tests" step above) is flaky, not clean,
+        # and its warning points here for e2e-cycle-1-logs.txt or
+        # e2e-deps-attempt-*-logs.txt. If this stayed failure()-only, that
+        # pointer would name a file nobody could ever retrieve. Do not
+        # simplify this back to failure() alone.
+        if: (failure() || env.E2E_RETRIED == 'true') && steps.should.outputs.run == 'true'
         uses: actions/upload-artifact@v7
         with:
           name: player-e2e-logs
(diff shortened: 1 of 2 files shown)


arsfeld:
## Measured: pull path

Run [33517128271 / 33517127415](https://github.com/getmydia/mydia/actions/runs/33517127415), `pull_request`, **success**.

This PR touches only `.github/workflows/ci-player-e2e.yml`, which is in the `relevant` filter but not in `server`, `toolbox` or `relay`, so all three images were pulled and no `docker buildx build` ran.

| | All-build baseline (33444861643) | This PR |
| --- | --- | --- |
| **Total job** | **27m08s** | **7m47s** |
| Free up disk space | 2m12s | skipped |
| Resolve server image | 12m30s (build) | 7s (pull) |
| Resolve toolbox and relay | 5m28s (build) | 1m21s (pull) |
| Run player E2E tests | 6m29s | 5m56s |

Better than the ~10m estimate in the description.

Two behaviours worth confirming from this run:

- `Free up disk space` correctly skipped. It is gated on the server building from source, and nothing built here.
- `Publish / Toolbox` correctly skipped, since it is `if: github.event_name == 'push'`.

The test step itself is unchanged at roughly 6 minutes, of which only about 2 minutes is actual test execution. The rest is `build_runner` and compiling the Flutter Linux app inside the container on every run. That is a separate ~3m30s opportunity, noted as out of scope.

**This run does not prove the cache fix.** No `docker buildx build` executed, so there is no cache manifest to import either way. That measurement needs the server-build path and follows in the next comment.

arsfeld:
## Measured: server-build path, and the cache fix is confirmed working

Run [33518165555](https://github.com/getmydia/mydia/actions/runs/33518165555), forced onto the server-build path by a throwaway commit touching `lib/mydia.ex` (reverted in the next commit). **Success.**

### The cache fix works

| Signal | Cold baseline (33444861643) | This run |
| --- | --- | --- |
| `importing cache manifest` | **0** | **2** |
| `CACHED` layers | **0** | **22** |

That is the whole point of the `Expose GitHub Actions runtime for BuildKit` step, and it is now doing something for the first time in this workflow's history.

### Timings

| | Cold baseline | This run |
| --- | --- | --- |
| **Total job** | **27m08s** | **21m00s** |
| Free up disk space | 2m12s | 2m03s |
| Resolve server image | 12m30s | **10m50s** |
| Resolve toolbox and relay | 5m28s | 1m26s |
| Run player E2E tests | 6m29s | 6m13s |

So a backend PR lands at about **21m**, against the roughly 19m estimated in the description. Honest result: the estimate was slightly optimistic.

### Why the server build only dropped 1m40s

The 22 cached layers are almost entirely the `flutter-builder` stage, which went from 3m07s to free. The Elixir builder stage barely cached at all, and one layer dominates everything:

| Layer | Cold | This run |
| --- | --- | --- |
| `mix compile` (699 `.ex` files, `mydia_p2p` + `mydia_subsync` in release mode, then 2 WASM plugins) | 7m51s | **8m11s** |
| `mix deps.compile` | 1m59s | 1m38s |

Part of that is expected and unavoidable: the throwaway commit changed `lib/`, which is what a backend PR does by definition, so `mix compile` must re-run.

But `mix deps.compile` re-ran too, and it should not have: `mix.exs` and `mix.lock` were untouched.

### The reason, and the highest-value follow-up

`Dockerfile:132` is `ENV BUILD_COMMIT=${BUILD_COMMIT}`, and it sits *before* `COPY mix.exs mix.lock` (138), `mix deps.get` (142), `mix deps.compile` (153) and `mix compile`.

Both `ci-docker.yml` and this workflow pass `BUILD_COMMIT=${{ github.sha }}`, which is different on every single commit. BuildKit folds the environment into the cache key of every `RUN` that inherits it, so **every `RUN` after line 132 has been permanently uncacheable across commits**, in this workflow, in `ci-docker.yml`, and in `release.yml`.

The run's own evidence fits precisely: `COPY mix.exs mix.lock` **cached** (a `COPY` does not read the environment), while every `RUN` after it re-ran.

Moving `ARG`/`ENV BUILD_COMMIT` down to just above `mix release`, the only consumer, would let `mix deps.get`, `mix deps.compile` and, on PRs that do not touch `lib/`, `mix compile` itself cache across commits. That is a handful of moved lines and it looks worth considerably more than the Rust `target/` work in the plan's follow-up list.

Not doing it here: this PR is scoped to the workflow file, and a Dockerfile change affects release builds and deserves its own PR and its own measurement.

### Follow-ups, in the order the measurements now justify

1. Move `ENV BUILD_COMMIT` below the dependency layers in the `Dockerfile`. Unblocks caching for three workflows at once.
2. Warm the Rust `target/` dir. Roughly 6m of `mix compile` is `mydia_p2p` and `mydia_subsync` compiling in release mode from cold, and no cache backend persists a `RUN --mount=type=cache`.
3. Cache the in-container `build_runner` and Flutter Linux compile, about 3m30s that repeats on every run.

arsfeld:
## CodeRabbit findings addressed

Both fixed in `ca3909dcc`.

**Pinned the runtime action to a SHA.** `crazy-max/ghaction-github-runtime@v3` became `@3cb05d89e1f492524af3d41a1c98c83bc3025124 # v3`, matching the SHA-pinning style already used in `ci.yml`. Worth doing here specifically because this action's entire job is to place `ACTIONS_RUNTIME_TOKEN` into the environment, so a moved tag would be handed a live token. SHA verified against `repos/crazy-max/ghaction-github-runtime/git/ref/tags/v3`.

**Fixed a real defect in the retry visibility.** This one was missed by every internal review and CodeRabbit was right to flag it. Two bugs:

- `start_deps()` succeeding on attempt 2 or 3 emitted nothing. The `e2e_relay` boot crash the retry exists for would recover silently and the job would look perfectly clean, which defeats the point of retrying rather than fixing.
- When test cycle 2 succeeded, the warning told the reader to see `e2e-cycle-1-logs.txt` in the artifacts, but `Upload logs on failure` was gated on `failure()` and the job had just succeeded. The artifact was never uploaded. The message pointed at a file nobody could retrieve.

Now both retry kinds emit a `::warning` and set `E2E_RETRIED=true` in `$GITHUB_ENV`, and the upload step is gated on `(failure() || env.E2E_RETRIED == 'true')` and renamed so its name is not a lie. `Collect service logs` stays `failure()`-gated.

Worth noting for anyone touching that condition later: a step whose `if:` contains no status-check function gets an implicit `success()`, which would have made this fix a no-op on the exact case it targets. The expression calls `failure()` directly, so no implicit `success()` is added and it fires correctly on a retried success.

## Unrelated CI failures

`Test` and `Test / PostgreSQL` are red on this branch. Both are pre-existing flakes, not caused by this PR, which changes one workflow file that `ci.yml` does not read:

- **SQLite:** a BEAM-internal `asmjit` JIT allocator assertion, exit 134 SIGABRT. The VM aborted mid-suite, so no test summary was printed.
- **PostgreSQL:** one failure, `CandidatePromotionTest` "separate database connections serialize competing promotions at ownership", a 10s `assert_receive` race already catalogued in `.github/ci-flakes.md`, which notes it recurring on diffs that touch only `.github/*.md`.

Both were independently reproduced on the two master runs immediately preceding this PR (`d16fda36f` and `9e883854d`).

arsfeld:
@coderabbitai review

PR #644, opened 2026-09-01
Flexible sidebar sections

arsfeld:
Answers the request in discussion #641 for an Anime category, without giving everyone a fixed Anime section.

## The shape

A **section** is a smart collection pinned to the sidebar. Nothing here is anime specific: `:anime_movie` and `:anime_series` already existed and are already assigned by `CategoryClassifier`, and `SmartRules` already accepted a `category in [...]` condition. What was missing was a way to promote a category slice into a first class library destination.

Sections render at `/sections/:id` through `MediaLive.Index` rather than the collection page, so a section gets the same toolbar as TV Shows: filters, sorting, selection, batch monitor, batch search, batch delete, batch edit. The seam is one option, `:base_query`, that lets the collection's rules supply the starting query.

A section marked **exclusive** claims its categories away from `/movies` and `/tv`, which is what actually answers the reporter's complaint that anime clutters the TV list. Only a rule that is exactly one `category in [...]` condition qualifies, because anything richer cannot be cleanly subtracted. Exclusivity is derived from the rules on every read, never stored, so it cannot drift.

Pins are per user. A shared collection pinned by its owner does not appear in anyone else's sidebar, so a section can never become a fixed section for the whole instance. Someone with no anime sees none of this.

## Getting one

Three presets (Anime, Cartoons, Documentaries) at `/sections/new`, or a **Pin to sidebar** button on any smart collection you own. When a library has at least 10 anime items and nothing claims them, the Movies and TV pages offer to create the Anime section once, dismissibly.

The preset list is limited to what `SmartRules` can express today. There is no 4K preset because there is no quality field, and no Recently Added preset because a smart collection requires at least one condition and a sort is not one.

## Things worth a reviewer's attention

- **NULL categories are kept deliberately.** SQL evaluates `NULL NOT IN ('anime_series')` to NULL, which `WHERE` treats as false. Without the explicit `is_nil` branch every item the classifier has not reached yet would vanish from the page it belongs on.
- **Exclusion is not applied on a section's own page.** A section's query already filters `category in [...]`; also applying `category not in [...]` would empty it out.
- **Counting goes through one aggregate.** `count_media_items/1` is a database aggregate that `count_movies/1` and `count_tv_shows/1` delegate to, so the sidebar badge and the page it links to share one exclusion set and cannot disagree. The Dashboard was reassigning those same assign keys with unexcluded counts and now uses the same exclusion.
- **Manual collections cannot be pinned.** A manual collection has `smart_rules: nil`, which falls back to `"{}"` and builds an unfiltered query, so pinning one would have put the entire library behind a section link. Refused in the context and again on the route.
- **Recovery paths.** Misclassification is the real risk of exclusivity. `category_override`, the per item category editor and `batch_reclassify` all already exist; the Movies and TV pages now name the section holding the missing items and link to it, and unpinning restores everything.

## Verification

`mix precommit` clean: compile, unused deps, format, Credo, and 10000 tests with 0 failures.

Migration adds three nullable columns to `collections` and is portable across SQLite and PostgreSQL.

## Summary by CodeRabbit

* **New Features**
  * Added customizable sidebar sections with Anime, Cartoons, and Documentaries presets.
  * Pin, unpin, rename, customize icons, and configure exclusive sections.
  * Browse dedicated section pages with smart filtering.
  * Added an Anime section suggestion that can be accepted or dismissed.
* **Bug Fixes**
  * Movie and TV counts now remain consistent when sections exclude categories.
  * Unclassified media remains visible when category filters are applied.
* **Documentation**
  * Documented sidebar sections and category exclusion behavior.

### DIFF
--- a/lib/mydia/accounts.ex
+++ b/lib/mydia/accounts.ex
@@ -16,6 +16,7 @@ defmodule Mydia.Accounts do
   alias Mydia.Accounts.{User, ApiKey, UserPreference, ApiKeyRateLimiter}
 
   @changelog_key "last_seen_changelog_version"
+  @anime_nudge_key "anime_nudge_dismissed"
 
   ## Users
 
@@ -499,6 +500,26 @@ defmodule Mydia.Accounts do
           _ -> true
         end
     end
+  end
+
+  @doc """
+  Returns true when the user has dismissed the offer to create an Anime section.
+  """
+  @spec anime_nudge_dismissed?(User.t()) :: boolean()
+  def anime_nudge_dismissed?(%User{} = user) do
+    pref = get_user_preference!(user)
+    Map.get(pref.preferences || %{}, @anime_nudge_key, false) == true
+  end
+
+  @doc """
+  Records that the user dismissed the Anime section offer. Permanent.
+  """
+  @spec dismiss_anime_nudge(User.t()) ::
+          {:ok, UserPreference.t()} | {:error, Ecto.Changeset.t()}
+  def dismiss_anime_nudge(%User{} = user) do
+    user
+    |> get_user_preference!()
+    |> update_preference(%{@anime_nudge_key => true})
   end
 
   ## API Keys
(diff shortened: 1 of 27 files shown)


arsfeld:
@coderabbitai review

PR #649, opened 2026-09-01
feat(discover): consolidate the add and request pages into Discover

arsfeld:
Makes Discover the single place to find and add media, and deletes the separate Add and Request pages.

From [discussion #642](https://github.com/getmydia/mydia/discussions/642#discussioncomment-18233438), where t00n1x pointed out the obvious thing: you go to Discover when you want to add something, find no search there, and have to detour through the Dashboard.

## Two bugs this fixes

Exploring the code turned up more than a navigation problem. Four separate code paths answered "what settings does this add use?" differently, and two of them were outright broken:

- **Adding from Discover never queued an indexer search.** `maybe_queue_search/2` lived only in the Add page, gated on a per-click button value. Adding the same title from Discover was silent.
- **`media.auto_search_on_add` was dead config.** Declared in the schema, loadable from `AUTO_SEARCH_ON_ADD`, documented, and read by nothing in `lib/`.

Both are fixed by giving every add path one resolver.

## What changed

**`Mydia.Media.AddDefaults`** resolves library, quality profile, monitored, season monitoring and search-on-add through four layers: an explicit per-item choice, then the user's preference, then the instance setting, then a hard fallback. A stale preference pointing at a deleted quality profile, or a library whose type no longer fits, degrades to the instance default rather than failing the add.

**Per-user add defaults** in Profile. Every field is tri-state, so even the booleans are three-option selects: a value, or "Use server default". A checkbox cannot express inherit.

**Discover gains** the search bar its backend already supported (previously reachable only by hand-typing `?q=`), a hide-in-library toggle, and the configure-before-adding modal, now hanging off the card caret so the primary button stays a one-click add.

**`Mydia.Media.AttrsFromMetadata`** lifts two private builders out of `SearchLive`, where they were untestable inside a 1500-line LiveView, and gives them coverage.

**Deleted:** `AddMediaLive` and `RequestMediaLive`, four routes, and 21 link sites repointed.

## Behaviour changes worth knowing

- **Guest request notes are gone.** The deleted Request page had a `requester_notes` field; the one-click Request button does not. The column and stored data are untouched, but the UI that displayed it is removed.
- **A stale `library_path_id` no longer flashes an error.** It falls back to the instance default silently, extending the resolver's "ignored rather than fatal" tolerance to a raw client param.
- **Guests lost their dedicated nav entries.** Discover is in the nav for every role and is now the request destination, so two entries pointing at the same place would both highlight. `/requests` is unchanged.

## One regression caught and fixed here

Deleting `/request/series` removed the only reachable path that stored a TVDB id in the right column. `handle_request_media/3` always wrote `tmdb_id`, so on a TVDB-configured instance a guest's TV request would resolve to the wrong title on approval. The bug predates this branch, but navigation always sent guests to the working page, so the deletion would have turned it live. Fixed in `handle_request_media/3`, `request_status_map/0` and both duplicate checks.

Tagging TVDB entries as `{:tvdb, id}` also closes a keyspace collision where a TMDB movie and a TVDB show sharing a numeric id could read as the same request.

## Testing

Full suite green: 13 doctests, 9996 tests, 0 failures.

## Known follow-ups

- `franchise_events.ex:206` reads `request_status_map/0` with a bare-key `Map.get` rather than `enrich_with_request_status/2`. Correct today only because franchise entries are always TMDB movies.
- `assets/populate-media.js` was repointed at Discover and verified against the markup, but not exercised against a running server.

## Summary by CodeRabbit

* **New Features**
  * Added search and “Hide owned” options to Discover.
  * Added configure-before-adding controls and profile-level defaults with server-default inheritance.
  * Unified movie and TV add/request actions through Discover.
  * Added configurable default season monitoring.
  * Added optional automatic searches after media is added.
  * Added a dismissible Player banner.
* **Bug Fixes**
  * Improved duplicate detection for TV shows and validation of library and quality selections.
  * Guests now have a streamlined request experience without requester notes.
  * Improved modal behavior when multiple configuration dialogs are open.


arsfeld:
Merged `master` in and worked through the review findings.

**Fixed**

- The `search_on_add` checkbox carried no `value` attribute, so a checked toggle submitted `"on"` and `params["search_on_add"] == "true"` read it as false: configuring an add with search enabled queued nothing. Now mirrors the `monitored` toggle's hidden-false plus `value="true"` pair. Every existing test drove the submit through explicit param maps, which bypass DOM serialization, so none of them could see it; the new test uses `form/2`.
- `AttrsFromMetadata` always resolved with a nil user, so every add through SearchLive ignored the acting user's library, quality profile and monitoring overrides. `:user` now threads through both builders and all four call sites.
- `library_path_exists_as_type?/2` accepted a disabled library, so a preference saved while one was enabled could still file an add into a library hidden from the UI. Now excludes disabled paths, matching `default_library_for/1`. `monitored` is deliberately still not required.
- A rejected `discover_hide_owned` write left the toggle flipped for the session and reverted on the next mount. Now flashes and leaves the value alone, matching `GridDensity.put/2`.
- The loader test always deleted `DEFAULT_SEASON_MONITORING` on exit, so a deployment setting it lost it to the test run. Moved into the shared capture-and-restore list.

**Not taken**

- Partial unique indexes on pending request ids. The check-then-insert race is real and predates this branch, but closing it needs a migration on both adapters for a duplicate row that is an operator-visible annoyance rather than corruption.
- Preserving a stale override when an unrelated field changes in the profile form. The select already renders "Use server default" for an id it cannot resolve, so persisting the empty value makes the stored preference match what the user is looking at; keeping a hidden stale id would make the UI disagree with the stored state.

**One regression the merge surfaced**

`Test / E2E Browser` failed reporting *21 tests, 0 failures* — the network guard, not a test, set the exit code, on `GET https://relay.mydia.dev/tmdb/genre/movie`. `Mydia.Metadata.genres/1` calls `Provider.Relay.fetch_genres/2` directly and never consults `Provider.Registry`, so `setup_metadata_stub` could not intercept it. Routing guests to Discover is what first put `guest_test.exs` on a page that fires `:load_genres` on mount. Fixed with `warm_genre_cache/2`, seeded after the stub setup since that clears the cache on its way in.

Local suite: 13 doctests, 10220 tests, 0 failures.

PR #658, opened 2026-09-02
fix: stop re-scan trashing media that is still on disk

### LINKED ISSUE #653: Re-scan files physically trashes media that is still on disk
## Impact

"Re-scan files" on a media item can physically move files that are still on disk into `.mydia-trash/`, leaving the item with no playable copy. Reported for both movies and TV shows. The toast reads:

> Re-scan complete! Found 0 new file(s), moved X file(s) to trash, refreshed metadata for 0 file(s)

## Reported behaviour

1. A monitored movie or show lists the same filename twice under Media Files, with only one physical file in the folder.
2. "View file details" on those rows errors out.
3. "Re-scan files" reports 0 new, X trashed, 0 refreshed, and the real files leave the library directory.
4. The same item set to Unmonitored re-scans correctly: nothing is trashed, metadata refreshes.

## Where it goes wrong

`rescan_movie/1`, `rescan_series/1` and `rescan_season/2` in `lib/mydia/library.ex` decide a file is gone from disk by scanning **one inferred directory** and diffing string paths, then hand every non-matching row to `Library.trash_media_file/1`:

* `lib/mydia/library.ex:1575` (movie), `:1204` (series), `:1372` (season) build `missing_files` as "rows whose `MediaFile.absolute_path/1` is not in the set of paths returned by scanning the base directory".
* `lib/mydia/library.ex:1583`, `:1212`, `:1397` trash each of those rows.
* `Mydia.Library.TrashStore.store/1` checks `File.exists?(source)` and, when the file *is* there, moves the bytes out of the library. So a row that the diff calls missing but that actually points at a real file gets the real file moved.

There is no per-file `File.exists?/1` check anywhere in the re-scan path. The only existence test is the one inside TrashStore, and by then the decision has already been made.

The worse branch is `{:error, :not_found}` at `lib/mydia/library.ex:1612` (movie), `:1261` (series), `:1450` (season). If the inferred base directory does not resolve, the code trashes **every** file belonging to the item, again with no per-file check. Any file that still exists is physically moved. An unmounted share, a renamed folder, or a base directory inferred from the wrong row all land here.

The base directory itself is inferred, not stored. `find_movie_base_directory/2` (`lib/mydia/library.ex:1789`) takes `Path.dirname/1` of each row's absolute path and picks the most frequent one; `find_series_base_directory/2` (`:1676`) does the same with the first path component. One bad row is enough to move the vote, and on a tie `Enum.max_by/2` just takes the first.

## Why the duplicate rows exist

`media_files` has **no uniqueness constraint on `(library_path_id, relative_path)`**. The original `unique_index(:media_files, [:path])` was dropped when `path` was superseded (`priv/repo/migrations/20251115185757_make_media_files_path_nullable.exs`) and was deliberately not restored (`priv/repo/migrations/20260817143147_add_import_group_id_to_media_files.exs:7`). `import_candidates` did get one on 2026-08-30 (`20260830143721_split_import_candidates_from_media_files.exs:51`); `media_files` never did.

`Library.get_media_file_by_path/2` resolves a path to `(library_path_id, relative_path)` via `calculate_relative_path/2`, so anything that resolves the same physical file through a different library path (overlapping library roots, a bind mount visible under two prefixes, a symlink) misses the lookup and inserts a second row. Two rows, same basename, different stored paths, one file on disk. That is exactly the state that makes the re-scan diff trash a live file.

The monitored/unmonitored split fits this: monitored items get their files placed by the import pipeline (`Mydia.Jobs.MediaImport`), unmonitored ones are typically only ever scanner-discovered, so only monitored items end up with rows from two different producers.

That last part is inference, not confirmed against a reproducing database. See "What would confirm it" below.

## What would confirm it

From the reporting instance, against `media_files`:

```sql
SELECT mf.id, mf.media_item_id, mf.episode_id, mf.library_path_id,
       lp.path AS library_root, mf.relative_path, mf.trashed_at
FROM media_files mf
LEFT JOIN library_paths lp ON lp.id = mf.library_path_id
WHERE mf.media_item_id = '<the affected item id>'
   OR mf.episode_id IN (SELECT id FROM episodes WHERE media_item_id = '<the affected item id>')
ORDER BY mf.relative_path;
```

Two rows with the same basename and different `library_root` or `relative_path` confirm the mechanism above. Also useful: the `Re-scanning movie` / `Re-scanning TV series` log line, which records the `directory` the re-scan picked.

The "View file details" error should come with a stacktrace worth having: `show_file_details/2` (`lib/mydia_web/live/media_live/show/file_events.ex:378`) calls `Library.get_media_file!/1` with no preload, so `library_path` reaches the modal unloaded.

## Proposed direction

1. Never trash a row whose file is present. Guard the trash call in all three re-scan paths on `File.exists?/1` for that specific row, not on the directory diff.
2. Drop the `{:error, :not_found}` mass-trash branches. A base directory that does not resolve is a scan failure, not evidence that the library is empty. Report it and change nothing.
3. Add a unique index on `media_files (library_path_id, relative_path)` where `trashed_at IS NULL`, with a data migration that folds existing duplicates.
4. Preload `library_path` in `show_file_details/2`.
5. Regression tests: a re-scan whose base directory is missing must trash nothing; a re-scan with a second row pointing outside the scanned directory must trash nothing while the file is on disk.

Items 1 and 2 stop the data loss on their own and are worth shipping ahead of the rest.

arsfeld:
Closes #653.

"Re-scan files" decided a media file was gone by scanning one inferred directory and diffing path strings, then handed every unmatched row to `trash_media_file/1`. That calls `TrashStore.store/1`, which checks `File.exists?` and, finding the file present, physically moves the bytes out of the library. Operators lost media, and the toast reported a clean sweep while doing it.

Three separate conditions reached that code with a false "missing" verdict: a base directory chosen by voting on `Path.dirname` across the item's rows, a scan whose own `errors` list was never read, and an explicit branch that trashed every file for an item whenever the directory would not resolve.

## The fixes

- A shared `reject_files_still_on_disk/1` guard in `Mydia.Library`, called by all three re-scan paths. A row the directory diff calls missing but whose own file is on disk is logged and skipped, never trashed. This closes the data loss regardless of which condition produced the false verdict.
- The three mass-trash-on-missing-directory branches now return `{:error, :scan_failed}` and write nothing. A directory that will not resolve is a failed scan, not evidence the library is empty. All three async handlers already had catch-all error clauses, so this surfaces as a failure flash rather than a success toast.
- `scan_errors` is carried out of the re-scan functions and into the toast, so a scan that could not read the files stops reporting "Found 0 new file(s)" with no further explanation.
- A migration folds duplicate `media_files` rows onto one row per physical file, grouping by canonical absolute path so duplicates reached through different library paths (a bind mount, a symlink, a trailing slash) fold together. Dependents across seven tables are repointed onto the survivor, with each guarded update followed by an explicit delete rather than relying on `ON DELETE CASCADE`, since some connection setups run with `PRAGMA foreign_keys=off`. Losing rows are deleted, not soft-trashed: `purge_old_trashed_media_files/1` treats a trashed row with no trash metadata as legacy and deletes the file at its library path once retention expires, so soft-trashing would have scheduled real files for deletion 30 days out.
- Rows whose library root is not absolute are excluded from the canonical fold, because `Path.expand/1` would otherwise resolve them against the working directory, and are folded separately on their literal `(library_path_id, relative_path)` pair. Since every row sharing a `library_path_id` shares one root and therefore one classification, the two folds partition cleanly and jointly guarantee the unique index can always be created. A migration that cannot create its index is a boot failure on every install, because `application.ex` starts `Ecto.Migrator` in the supervision tree.
- A partial unique index on `media_files (library_path_id, relative_path) WHERE trashed_at IS NULL`, with matching changeset constraints. `media_files` lost its uniqueness when `path` was superseded and never got it back, while the sibling `import_candidates` table gained one; that gap is what let two rows describe one file.
- The file details modal called `Jason.encode!` on a `FileMetadata` struct with no `Jason.Encoder` implementation, so it raised for every analyzed file. `FileMetadata` and `StreamInfo` now derive the protocol, and `show_file_details/2` preloads `library_path` so the modal shows the full path instead of falling back to the relative one.
- `media_import.ex` adopts an existing row when it hits the new constraint, instead of reporting `:database_error` and retrying. Re-running an import, or retrying after a partial failure, should reuse the row already there; the file already did this via `get_media_file_by_path/2` and the `{:existing, path}` conflict branch, and the constraint violation now lands in the same place.

## Notes

The prune `:duplicate_registration` check and the import-run duplicate-tolerance path were both kept rather than deleted. The index is partial in two ways: it does not cover rows with a NULL `library_path_id`, and it excludes trashed rows. Both guarded states remain reachable, through two orphaned rows in the first case and one active plus one trashed row at the same path in the second.

`priv/repo/README.md` gains a note on an `ecto_sqlite3` trap found here: a violation of a custom-named partial unique index is reported under a convention-derived name, so a `unique_constraint` keyed only to the real name never fires on SQLite and the insert raises instead of returning a changeset error. It has to be declared under both names.

`mix precommit` passes: 10082 tests, 0 failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate media imports from creating conflicting records; existing active files are reused safely.
  * Rescans no longer trash files when scanning fails or when files remain on disk but are outside the scanned directory.
  * Rescans now report files that could not be read instead of silently ignoring them.
  * Improved handling of duplicate and trashed media-file paths.

* **Improvements**
  * File details now display the complete absolute file path and expanded metadata in the details modal.
  * Improved reliability when importing files concurrently.

### DIFF
--- a/lib/mydia/jobs/media_import.ex
+++ b/lib/mydia/jobs/media_import.ex
@@ -1761,27 +1761,42 @@ defmodule Mydia.Jobs.MediaImport do
         {:ok, media_file}
 
       {:error, changeset} ->
-        # Check if this is a library type mismatch error
-        if has_library_type_mismatch_error?(changeset) do
-          media_type = if episode, do: "TV show", else: "movie"
+        cond do
+          # A concurrent import (this job racing its own retry, or two
+          # workers racing the same season pack) can lose the create/1 race
+          # after this process already decided the path was free: another
+          # insert won and the unique index on (library_path_id,
+          # relative_path) for active rows (migration 20260901234336) turns
+          # the loser's insert into a changeset error instead of a second
+          # row. Adopt the row that won instead of failing the import --
+          # the same "someone already put this here" reuse this function
+          # already does before attempting to place the file (see "Reusing
+          # existing media file" above) and the `{:existing, path}` branch
+          # of resolve_conflict_path/2, just caught one step later.
+          duplicate_active_path_error?(changeset) ->
+            adopt_duplicate_active_path(library_path.id, relative_path, path)
 
-          Logger.error("Library type mismatch during import",
-            path: path,
-            media_type: media_type,
-            download_id: download.id,
-            media_item_id: download.media_item_id,
-            episode_id: episode && episode.id,
-            errors: format_changeset_errors(changeset)
-          )
+          has_library_type_mismatch_error?(changeset) ->
+            media_type = if episode, do: "TV show", else: "movie"
 
-          {:error, :library_type_mismatch}
-        else
-          Logger.error("Failed to create media file record",
-            path: path,
-            errors: inspect(changeset.errors)
-          )
+            Logger.error("Library type mismatch during import",
+              path: path,
+              media_type: media_type,
+              download_id: download.id,
+              media_item_id: download.media_item_id,
+              episode_id: episode && episode.id,
+              errors: format_changeset_errors(changeset)
+            )
 
-          {:error, :database_error}
+            {:error, :library_type_mismatch}
+
+          true ->
+            Logger.error("Failed to create media file record",
+              path: path,
+              errors: inspect(changeset.errors)
+            )
+
+            {:error, :database_error}
         end
     end
   end
@@ -1891,6 +1906,42 @@ defmodule Mydia.Jobs.MediaImport do
   defp maybe_put(map, _key, nil), do: map
   defp maybe_put(map, _key, ""), do: map
   defp maybe_put(map, key, value), do: Map.put(map, key, value)
+
+  # Matches the specific violation `guard_unique_active_path/1` in
+  # MediaFile attaches to :relative_path (declared under two constraint
+  # names there because ecto_sqlite3 reports a convention-derived name
+  # rather than the real one for a partial index) -- never a genuine
+  # validation failure, which must still surface as an error.
+  defp duplicate_active_path_error?(changeset) do
+    Enum.any?(changeset.errors, fn {field, {_message, opts}} ->
+      field == :relative_path and Keyword.get(opts, :constraint) == :unique
+    end)
+  end
+
+  # Looks up the row that won the race and adopts it. A miss here (the
+  # constraint fired but no active row exists for this path) means it was
+  # trashed or deleted between the insert and this read -- nothing to adopt,
+  # so report the original failure and let the caller retry.
+  defp adopt_duplicate_active_path(library_path_id, relative_path, path) do
+    case Library.get_media_file_by_relative_path(library_path_id, relative_path) do
+      nil ->
+        Logger.error("Duplicate-path constraint hit but no active row found to adopt",
+          path: path,
+          library_path_id: library_path_id,
+          relative_path: relative_path
+        )
+
+        {:error, :database_error}
+
+      existing ->
+        Logger.info("Reusing existing media file after a duplicate-path race",
+          path: path,
+          id: existing.id
+        )
+
+        {:ok, existing}
+    end
+  end
 
   # Checks if a changeset has a library type mismatch error
   defp has_library_type_mismatch_error?(changeset) do
--- a/lib/mydia/library.ex
+++ b/lib/mydia/library.ex
@@ -651,13 +651,22 @@ defmodule Mydia.Library do
   present could not be moved. A row marked trashed while its file sits in the
   library is precisely the inconsistency this move exists to remove, so the two
   are kept in step or neither changes.
+
+  ## Options
+
+  Passed straight through to `Mydia.Library.TrashStore.store/2`; see its docs.
+  Notably `:move` (default `true`) - `move: false` refuses to move a file that
+  is present on disk instead of trashing it, leaving the row untouched. The
+  three re-scan functions in this module pass it: a file the diff called
+  missing but that has since reappeared must never be moved out of the
+  library.
   """
-  @spec trash_media_file(MediaFile.t()) ::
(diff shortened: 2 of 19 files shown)


arsfeld:
Both CodeRabbit findings were real and are fixed; the re-posted copies above are stale, re-anchored against the updated diff but carrying the original text (the first still cites line 705 and `TrashStore.store/1`).

**Self-referencing supersedes pointer** (5fafd1d9d). Confirmed dangerous before fixing: with a self-reference, `Upgrades.superseded_file/1` returns the row itself, `finalize_comparison/2` compares a file against itself, and the upgrade branch can trash the file it is keeping. The repoint now excludes the winner and clears its pointer instead:

```sql
UPDATE media_files SET supersedes_media_file_id = <winner>
WHERE supersedes_media_file_id = <loser> AND id <> <winner>;

UPDATE media_files SET supersedes_media_file_id = NULL
WHERE id = <winner> AND supersedes_media_file_id = <loser>;
```

Covered by a test seeding a winner that supersedes the loser it absorbs.

**Check-to-trash race** (882377cfb). Closed structurally rather than narrowed. `TrashStore.store/2` takes a `move` option, default `true` so every other caller is unchanged. With `move: false` a present file returns `{:error, :file_present}` and `do_store/2`, the only function in that module reaching `File.rename/2` or `File.cp/2`, is never called. All three re-scan sites pass `move: false`, so a file that reappears between the guard and the trash leaves the row active and the bytes in place. `reject_files_still_on_disk/1` stays as the operator-facing warning above it.

`mix precommit` passes at 10087 tests, 0 failures, and CI is green including the PostgreSQL leg.

PR #662, opened 2026-09-02
feat(ui): hold the grid shape while poster grids load

arsfeld:
Three poster grids fetch metadata asynchronously and showed a centered spinner while waiting, which collapses the layout and re-expands it when results land.

The worst case is not the initial load. `discover_live/index.ex` sets `@loading` on every filter, genre, year, rating, sort and media-type change, not only on mount. On a filter change the grid is already on screen at full height, and today it collapses to a spinner and re-expands. A placeholder that holds the shape removes that.

## What changed

`MydiaWeb.PosterCardComponents.poster_card_grid_skeleton/1` renders N placeholder cards shaped like a real poster card, into a caller-supplied column class. It sits directly below `poster_card_body/1`, the thing it imitates, so the two shapes stay visible to each other in review.

Wired into three loading branches:

| Site | Columns | Count |
| --- | --- | --- |
| Discover results grid | `grid_columns_class(@grid_density)`, all three densities | 20, an uncapped provider page |
| Dashboard Trending Movies | fixed 2/3/4/5 | `@trending_rail_limit` |
| Dashboard Trending TV | fixed 2/3/4/5 | `@trending_rail_limit` |

`columns` is a required attr with no default, because the two consumers genuinely differ and a default would let a call site silently lay out differently from its own results grid.

## Height parity is measured, not assumed

The premise is that a placeholder reserves the same height as a real card, so the bar heights were measured in headless Chromium rather than derived from reading CSS.

The first measurement found the placeholder 12px short in both themes. The cause decomposed exactly: Tailwind's `text-xs` line-height is `1rem`/16px and is not derived from its 0.75rem font-size (4px), and every real action element carries its own `mt-2` on top of `card-body`'s gap (8px). After correcting `h-3` to `h-4` and `gap-2` to `gap-4`, a placeholder and a real card both measure 341.5px, for two-line and one-line titles, in light and dark.

`test/mydia_web/features/discover_skeleton_height_parity_test.exs` makes that permanent. It uses a new `assert_same_height/4` in the existing `MydiaWeb.FeatureCase.Geometry`, and avoids racing the transient skeleton by fetching Discover's disconnected render (which always contains the skeleton, since `@loading` is set outside any `connected?/1` check) and splicing it beside the real grid, so both cards are measured in one document at one instant. The guard was verified to fail: perturbing the injected card by 12px produces `#injected-skeleton-grid .card is 331.5px but #discover-grid .card is 343.5px (delta 12px, tolerance 1px)`.

## Dashboard count coupling

The rails cap their data with `Enum.take/2`, so a skeleton at the component's default of 20 would have shown 4 rows collapsing to 2. Both the cap and the placeholder count now read one module attribute, `@trending_rail_limit`, and the regression test asserts against `DashboardLive.Index.trending_rail_limit()` rather than a literal, so the two cannot drift apart while staying green.

## Deliberately out of scope

- **Search.** Its loading state already shows per-indexer progress and streams results into a table, so grey rows would replace information with less information.
- **Discover's `loading_more`.** The grid above it is already populated and holding its shape.
- **Per-poster image placeholders**, and **the main library grid**, which loads synchronously and would need an async refactor first.
- **`assign_async`.** All 12 async sites in the tree hand-roll a boolean assign; introducing a second idiom for three of them was rejected.

## Accessibility

The replaced Discover spinner carried visible text ("Loading movies..."). The grid wrapper carries `role="status"` and `aria-label="Loading titles"` so that is not a regression for assistive tech. Reduced motion needs no handling: daisyUI gates the shimmer behind `prefers-reduced-motion: no-preference`.

## Verification

`mix precommit` green: 10102 tests, 0 failures, plus `credo --strict`. Feature tests pass under `./dev feature-test`.

One full run failed on `CandidatePromotionTest` with `database is locked`, the SQLite load-contention flake catalogued in `.github/ci-flakes.md`. It passes 10/10 in isolation and three of four full runs on this branch were green. Per that catalogue's own instruction, a second run on the identical tree was used to confirm.

## Summary by CodeRabbit

- **New Features**
  - Added responsive placeholder card grids for Dashboard trending movies and TV sections while content loads.
  - Added matching loading placeholders to Discover results, preserving the selected grid density and layout.
  - Loading placeholders now mirror the size and structure of finished poster cards.

- **Bug Fixes**
  - Improved loading-state consistency by aligning placeholder card counts and dimensions with the content that follows.
  - Replaced centered loading spinners with representative content skeletons for a smoother browsing experience.

### DIFF
--- a/lib/mydia_web/components/discover_components.ex
+++ b/lib/mydia_web/components/discover_components.ex
@@ -15,6 +15,51 @@ defmodule MydiaWeb.DiscoverComponents do
     router: MydiaWeb.Router,
     statics: MydiaWeb.static_paths()
 
+  # A TMDB page. Both Metadata.trending_movies/0 and Metadata.discover/2 return
+  # a provider page uncapped, so a skeleton grid of this size is replaced
+  # one-for-one by real cards and the grid does not resize when they land.
+  @skeleton_count 20
+
+  @doc """
+  Renders a grid of placeholder cards shaped like `trending_card/1`.
+
+  Sits beside `trending_card/1` rather than in a generic module because the two
+  have to keep the same shape, and co-location is the only thing that makes a
+  drift visible in review.
+
+  `columns` is a parameter because the consumers genuinely differ: Discover
+  drives columns from `GridDensityComponents.grid_columns_class/1` across three
+  density levels, while the Dashboard trending rows hardcode a 2/3/4/5 grid.
+  Each caller passes the same string its real grid uses, so the placeholder and
+  the results lay out identically.
+
+  Reduced motion needs no handling here. daisyUI gates the shimmer behind
+  `prefers-reduced-motion: no-preference` and leaves a flat base-300 fill
+  otherwise.
+  """
+  attr :id, :string, default: nil
+  attr :count, :integer, default: @skeleton_count
+  attr :columns, :string, required: true
+  attr :gap, :string, default: "gap-4 md:gap-5"
+
+  def poster_card_grid_skeleton(assigns) do
+    ~H"""
+    <div id={@id} class={["grid", @gap, @columns]} role="status" aria-label="Loading titles">
+      <div :for={_ <- 1..@count} class="card bg-base-100 shadow-sm">
+        <div class="skeleton aspect-[2/3] rounded-t-box rounded-b-none"></div>
+        <div class="card-body p-3">
+          <div class="skeleton h-4 w-full"></div>
+          <div class="skeleton h-4 w-2/3"></div>
+          <div class="mt-auto flex flex-col gap-2">
+            <div class="skeleton h-3 w-10"></div>
+            <div class="skeleton h-8 w-full"></div>
+          </div>
+        </div>
+      </div>
+    </div>
+    """
+  end
+
   @doc """
   Renders a trending media card with poster, status badge, title, year,
   and an action button (Add to Library / Request / Go to Movie/Show).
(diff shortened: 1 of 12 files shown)


arsfeld:
Both CodeRabbit findings reviewed. One fixed in `ccc3c2c94`, one acknowledged as a real gap I am not closing here, with reasoning.

### Fixed: delta precision in `assert_same_height/4`

Correct, and it mattered more than it looked. `Float.round(delta, 1)` rendered a failing `1.04px` as `1px`, so a genuine failure printed `delta 1px, tolerance 1px` — a message that reads like it should have passed, sending the reader after a bug in the assertion instead of in their layout. Now rounds to two places, with a self-test pinning it.

Making the message finer immediately broke two assertions that had been quietly matching the old rounding, including the parity guard's own deliberate-failure test, which pinned `delta 12px` while the real delta is `11.97px` (the card height is sub-pixel, so shrinking by a whole number of pixels does not yield a whole-number difference). Both now assert the delta's magnitude rather than its exact rendering, so they test our formatting rather than Chrome's device-pixel snapping.

### Not closing: independent rail completion

The finding is correct and I verified it. All four Dashboard tests observe both rails in the *same* state, so a wiring error pointing both skeletons at `@trending_movies_loading` would pass every one of them.

I am not closing it because it is not reachable through `Phoenix.LiveViewTest` at reasonable cost:

- Assigns cannot be set directly, so the two rails cannot be put into different states from a test.
- Both loads are queued during `mount/3` (`send(self(), :load_trending_movies)` and `:load_trending_tv`) and both drain before the first `render`, so there is no window where one is settled and the other is not.
- Leaving one rail's cache unwarmed does not help: the fetch fails and that rail settles with an empty list, giving all-settled again rather than one-loading.
- The dashboard template is not a function component, so there is no render-with-assigns seam either.

Closing this properly means restructuring how the rails load, which is out of scope for a loading-state change. What the existing tests do cover: both ids present while loading, both absent once settled, correct column classes, and the placeholder count tied to `DashboardLive.Index.trending_rail_limit()` so the cap and the placeholder cannot drift apart.

Recorded rather than silently dropped. Happy to take a suggestion if there is a seam I have missed.

PR #665, opened 2026-09-02
fix(import-lists): close the audit findings and enable by default

arsfeld:
An audit of the import lists feature found that `ENABLE_IMPORT_LISTS` gated nothing but a nav link, and turned up eight bugs behind it. This fixes all of them, makes the flag real, and turns the feature on by default.

## The flag gated a nav link

`FeatureFlags.enabled?/0` had exactly one call site in the app, the sidebar link in `layouts.ex`. The route, the every-15-minutes scheduler cron, and the sync and auto-add jobs all ran regardless, so an operator who set `ENABLE_IMPORT_LISTS=false` still had the feature running on a schedule and reachable by URL.

The scheduler now returns early when disabled, following the `CardigannHealthCheck` precedent already in the repo, and the LiveView redirects on mount. With the flag actually gating the feature, the default flips to on: syncing an external list is a normal admin action and should not require setting an environment variable first. An explicitly configured `false` still wins; only a genuinely missing key falls back to the new default.

## Bugs fixed

| # | Bug | Effect |
|---|---|---|
| 1 | `mark_sync_error/2` never touched `last_synced_at` | A failing list stayed permanently due, so the cron re-enqueued it every 15 minutes forever, roughly 96 relay hits a day per broken list |
| 2 | `build_endpoint/2` fell through to trending movies | An Upcoming + TV Shows list silently filled with movies stored as TV shows; auto-add then resolved movie TMDB IDs as series |
| 3 | `page: 1` hardcoded | Every list capped at about 20 items |
| 4 | Both upsert branches returned the same shape | The `updated` counter was unreachable, so every sync reported every item as new |
| 5 | Upsert was select-then-insert | Concurrent syncs raced the unique constraint and the loser was silently dropped |
| 6 | Duplicates marked skipped, then immediately added | The skip was erased, so stats and database disagreed and the Skipped filter never showed them |
| 7 | `check_duplicate` matched only `tmdb_id` | Library items from a filesystem scan with no external id were added a second time |
| 8 | Custom URL fetched with no scheme or destination check | An admin-supplied URL could reach loopback, the LAN, container neighbours, or a cloud metadata endpoint |

Notes on the two with the most design in them:

**Backoff (1).** Failures now record a timestamp and a consecutive-failure count and back off exponentially, capped at 24 hours, reset on success. State lives in the existing `config` map, so no migration. The exponent is clamped: `2 ** 1024` overflows a float and raises, and since `list_sync_due_lists/0` runs every list through `sync_due?/1`, one raise would have stopped syncing for all lists, not just the broken one.

**SSRF (8).** Requests are limited to `http`/`https` against public addresses, checked on the *resolved* IP rather than the host text, with IPv4-mapped IPv6 unwrapped and every redirect hop revalidated. Operators legitimately serving a list from their own NAS or container network can opt back in with `IMPORT_LISTS_ALLOW_PRIVATE_DESTINATIONS=true`. DNS rebinding remains out of scope and is documented as a deliberate trade-off in the moduledoc.

## UI

The auto-add checkbox now says what it does. Auto-added items are monitored by default, which hands them to the hourly `MovieSearch` and half-hourly `TVShowSearch` in `all_monitored` mode, so ticking it on a popular-movies list starts real downloads. Nothing in the UI said so. The Media Type select now only offers types the chosen source actually supports, and the form keeps the shown value and the submitted value in sync.

## Tests

The feature had 40 tests covering only context CRUD and changesets. Both providers, all three workers, the LiveView, and the entire add-to-library path had none, so the tested surface was the half that could not fail in a way that mattered.

Adds 81 tests, including the first LiveView test for the page: 121 across the feature, all green. Full suite green at 10,227.

## Summary by CodeRabbit

* **New Features**
  * Import Lists is enabled by default and presented as a stable feature.
  * TMDB imports now support multi-page lists.
  * Media type options adjust automatically based on the selected source.
  * Custom URL imports support configurable private-destination access.

* **Improvements**
  * Improved duplicate matching and linking of existing library items.
  * Added sync failure tracking with automatic retry backoff.
  * Added validation for incompatible list sources and media types.
  * Auto-add forms now warn about automatic downloads for monitored items.

* **Security**
  * Custom URL imports validate protocols, destinations, redirects, and response sizes.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -452,9 +452,12 @@ config :mydia, :features,
   # Can be overridden via ENABLE_CARDIGANN environment variable
   cardigann_enabled: true,
   # Enable/disable Import Lists feature
-  # When enabled, shows the Import Lists UI for syncing external lists (TMDB watchlists, etc.)
+  # When enabled, runs the Import Lists UI and scheduled sync for external
+  # lists (TMDB watchlists, etc.); when disabled, the admin route redirects
+  # away and the scheduler job is a no-op
+  # On by default: syncing an external list is a normal admin action
   # Can be overridden via ENABLE_IMPORT_LISTS environment variable
-  import_lists_enabled: false,
+  import_lists_enabled: true,
   # Enable/disable Remote Access feature (iroh-based peer-to-peer connectivity)
   # When enabled, starts the p2p server for remote device pairing and media streaming
   # On by default: connecting a player is a normal user action, and requiring an
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -250,9 +250,21 @@ if config_env() == :prod do
 
   import_lists_enabled =
     case System.get_env("ENABLE_IMPORT_LISTS") do
-      "true" -> true
-      "false" -> false
-      _ -> Application.get_env(:mydia, :features)[:import_lists_enabled] || false
+      "true" ->
+        true
+
+      "false" ->
+        false
+
+      _ ->
+        # An explicit `false` configured elsewhere must still win: only a
+        # genuinely missing key falls back to the compile-time default
+        # (true), so `Keyword.get/3` (which can't tell "missing" from
+        # "explicitly false") would be wrong here.
+        case Keyword.fetch(Application.get_env(:mydia, :features, []), :import_lists_enabled) do
+          {:ok, value} -> value
+          :error -> true
+        end
     end
 
   remote_access_enabled =
@@ -411,9 +423,21 @@ if config_env() in [:dev, :test] do
 
   import_lists_enabled =
     case System.get_env("ENABLE_IMPORT_LISTS") do
-      "true" -> true
-      "false" -> false
-      _ -> Application.get_env(:mydia, :features)[:import_lists_enabled] || false
+      "true" ->
+        true
+
+      "false" ->
+        false
+
+      _ ->
+        # An explicit `false` configured elsewhere must still win: only a
+        # genuinely missing key falls back to the compile-time default
+        # (true), so `Keyword.get/3` (which can't tell "missing" from
+        # "explicitly false") would be wrong here.
+        case Keyword.fetch(Application.get_env(:mydia, :features, []), :import_lists_enabled) do
+          {:ok, value} -> value
+          :error -> true
+        end
     end
 
   remote_access_enabled =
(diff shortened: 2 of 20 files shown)


arsfeld:
@coderabbitai review

arsfeld:
Thanks, three of the four were real. Fixed in 92319107c.

**1. `Repo.one/1` can raise on the title+year fallback — fixed.** Confirmed: `MediaItem` enforces uniqueness on `tmdb_id` and `tvdb_id` only, and NULLs do not collide, so two scanned copies of one title genuinely sit there. `Ecto.MultipleResultsError` would have propagated out of the sync job and the admin add action. Added `order_by: [asc: m.id], limit: 1`. The regression test asserts the call returns one of the duplicate rows and keeps returning the same one; it deliberately does not assert *which*, since the id is a UUID and ordering by it is stable but not insertion order.

**3. Non-global ranges accepted as public — fixed.** Added `198.18.0.0/15`, `fec0::/10` and `ff00::/8`, with literal-address regression tests for each plus the IPv4 multicast case.

**4. `async: true` with global env mutation — fixed.** Set the module to `async: false`. Worth noting for anyone reading later: `Mydia.DataCase` forces `async: false` on SQLite, so this only ever leaked on the PostgreSQL job, which is exactly the green-SQLite / red-Postgres split shape documented in `.github/ci-flakes.md`.

**2. DNS rebinding between validation and connect — acknowledged, not fixed.** This one is real and I am leaving it, deliberately rather than by oversight; it is called out under "Residual risk" in the module doc.

The reasoning: closing it means connecting to the validated IP while carrying the original hostname for `Host` and TLS SNI, and then re-doing that for every redirect hop. In Req 0.7.1 that is `connect_options: [hostname: ...]` per hop, and getting it wrong silently weakens certificate validation, which trades a difficult-to-exploit issue for an easier one. Against that, the threat model here is narrow: `list_url` is admin-only configuration, so an attacker needs to already control either the admin account or the DNS for a domain the admin chose to point Mydia at. An attacker with the former has far more direct options.

The guard is defence in depth on an admin-configured field, and the pre-connect check still closes the ordinary cases: literal internal addresses, hostnames that resolve to them, and redirects into them. If this moves to a user-supplied field, the calculus changes and pinning becomes worth the complexity.

PR #666, opened 2026-09-02
feat(trash): make the trash visible, restorable and accountable

arsfeld:
Trash was fully built on the backend and completely invisible. Files moved to `.mydia-trash` and were purged after 30 days, but nothing let an operator see what was in there, restore something wrongly trashed, reclaim the space early, or find out why a file was trashed. The only trace was the activity feed, which said it happened and offered nothing to do about it.

This adds `/admin/config/trash`, and closes the two loops that building it exposed.

## Why a file was trashed

`media_files` gains a nullable `trashed_reason` enum (`:missing`, `:upgraded`, `:upgrade_rejected`, `:pruned`, `:manual`), stamped at **all seven** call sites. Three of those live inside `Mydia.Library` itself, in the re-scan reconciliation paths, and are easy to miss because they do not read like producers; leaving them unstamped would have made every file trashed by an operator's own re-scan show as "Unknown". Rows trashed before the column existed stay nil and render as Unknown, with no backfill: the information to reconstruct them does not exist.

## The page

Newest first, paginated, with daisyUI filter chips per reason and per-row Restore and Delete permanently. Bulk selection runs through a new `Mydia.Jobs.TrashAction` Oban worker, because restoring hundreds of files moves bytes and must survive the operator closing the tab.

Selecting "all N matching" travels to the job as the **filter**, never as a list of ids. The page is paginated, so an id list would silently truncate the selection to the visible page, which is the exact failure the feature exists to prevent: a dropped NAS mount makes one scan trash the whole library, and recovery is filter to Missing, select all, Restore.

## Bytes nothing will ever purge

`TrashStore.audit/0` classifies every `<root>/<id>` container as tracked, retained, or orphaned. The obvious rule ("no `media_files` row") is wrong twice: it misses the commonest case, where a restore found the library path re-occupied and deliberately kept the trashed copy while the row went active, and it races an in-flight trash, which moves bytes before stamping `trashed_at`. `sweep/1` therefore acts only on entries `audit/0` produced and skips anything written in the last hour.

## Two loops this exposed

- **Manual delete now defaults to trash.** Every automated path already soft-deleted; only the human's own delete button was a hard `Repo.delete` plus `File.rm`, so the one deletion performed deliberately was the only one with no undo. It becomes a three-way choice, with permanent deletion still one click away and labelled plainly. This changes a destructive action's default on purpose.
- **Retention moved into the layered config** (`media.trash_retention_days`), settable via YAML, the DB overlay or `TRASH_RETENTION_DAYS`, instead of a compile-time key needing a restart.

## Notes for review

- `Library.restore_media_file/1` gains a `{:ok, file, :trash_copy_retained}` return so callers can tell a clean restore from one that had to leave a second copy behind. `LibraryScanner` was its only caller.
- `purge_old_trashed_media_files/1` changes `<` to `<=` on the cutoff. Both sides are truncated to the second, so "Empty trash now" (`days: 0`) previously spared anything trashed in the current second.
- TV rows have `media_item_id` NULL and hang the show off `episode.media_item`. The row label matches the episode clause first and preloads `episode: :media_item`; the tests use episode fixtures, since a movie fixture passes either way and that is how this bug class shipped twice before.
- `mix precommit` green: 10161 tests, 0 failures.

Known follow-ups, deliberately not in scope: "Select all N matching" only appears once one row is ticked, and retention has no settings-page form control (shared with its neighbours in the `media` section).

## Summary by CodeRabbit

* **New Features**
  * Added an admin Trash page to review, filter, restore, purge, and bulk-manage trashed media.
  * Added trash reason tracking, configurable retention, summaries, and orphaned-file auditing and cleanup.
  * File deletion now offers library-only, recoverable trash, or permanent deletion options.
  * Added a “Not this item” action to return incorrectly matched files for review.
  * Added configurable default season monitoring.

* **Bug Fixes**
  * Improved restoration handling when destination paths are occupied.
  * Improved large-library storage total calculations.
  * Trash retention can now be configured at runtime, including through environment settings.


arsfeld:
Merged `master` (three additive conflicts in the `media` config section, `default_season_monitoring` against `trash_retention_days`; both kept) and worked the five review findings in 43b3209.

Fixed as reported:

- **Page offset outran the result set.** `load/1` now counts, clamps, then queries.
- **Sweep ran in the LiveView process.** Moved to `start_async/3`. Dropping the audit as the task starts removes the button, so a second click has nothing to re-sweep.
- **`trash_retention_days` accepted an explicit null.** Now `validate_required`, so it fails at config load rather than a day later in `TrashCleanup` on `-nil`.
- **`purge_media_file/1` deleted by primary key.** The delete moved under the same `trashed_at IS NOT NULL` guard `purge_old_trashed_media_files/1` already used, so a restore landing between the fetch and the delete makes this a refusal instead of destroying the restored row.

Two places where I did not take the suggestion as written:

**Restore/purge serialization.** The guarded delete closes the half that loses data. The remaining window is `discard/2` removing the NFO sidecar of a file restored a moment earlier. Closing that properly needs a claim column plus a protocol restore participates in, and the lock-based version is not portable: SQLite is the default adapter here, so `FOR UPDATE` is not available. That is more than this PR should carry for a regenerated sidecar, now that the row and the media file survive either way.

**Audit classification.** Took the mismatch half: a trashed row whose recorded path lies in a different container is now `:orphaned`, since the purge deletes the recorded path and never `<root>/<id>`.

Did not take the missing-path half. Trashing moves the bytes before it stamps `trashed_at`, so a trashed row with no recorded path can be a trash still in flight, and the one-hour skip in `sweep/1` does not cover a stall longer than an hour. Reporting it would put a file inside its retention window behind a button labelled Sweep. The two errors are not symmetric: classifying debris as tracked leaves disk used, classifying a tracked container as orphaned deletes a recoverable file. That row stays `:tracked`, with a test pinning both halves.

Regression tests for all five.

PR #669, opened 2026-09-02
fix(downloads): stop the matcher attaching unrelated releases to library items

### LINKED ISSUE #653: Re-scan files physically trashes media that is still on disk

arsfeld:
Closes #653.

## The defect

`TorrentMatcher` scores a candidate library item against a release name with Jaro-Winkler similarity. That metric measures character overlap inside a sliding window, with a bonus for a shared prefix. It was built for typos in short strings, and it has no concept of a word, so two unrelated titles that merely start alike score high.

Two confirmed cases:

- The reporter's screenshot: `The Super Mario Galaxy Movie` attached to a movie called `Supergirl`, at **0.897** confidence.
- Production on galactica: five season packs of `New Amsterdam` attached to `The New Years` (2024, `Los años nuevos`), at **0.886**, with no `New Amsterdam` item in that library at all. None had imported yet, so nothing had moved.

Two gates stood between a release name and a file move, and both were fooled:

- `calculate_movie_confidence/2` scores `jaro_winkler * 0.7 + year`. With an exact year (+0.3) the effective bar on the title alone was **0.715**.
- `calculate_tv_show_confidence/2` is thinner still: bare Jaro-Winkler against 0.8, no year term and no penalties.
- The `TargetContext` wrong-show guard uses `Text.title_similarity/2`, which falls through to plain Jaro and demotes only below 0.5.

The two gates share a blind spot, so they are effectively one gate. Worse, the guard has an upward branch: it adds `binding_confidence * 0.05` to anything already over the threshold. That is what took the production match from 0.851 to 0.886. A guard whose only job is demotion was making a wrong match look better evidenced than it was.

## The fix

**A token-coverage gate before scoring.** `Text.title_token_coverage/2` returns the fraction of the *library* title's significant words that appear in the release title. A candidate below 0.75 is rejected before it is scored, on all three title-match paths.

The measure is one-directional, and that asymmetry is the point. Extra words on the release side are normal: subtitles, edition markers, cruft the parser did not strip. A word missing from the release side means a different work. A symmetric measure cannot express this. F1 was tried first and the existing suite refuted it: `Dr. Strangelove` against its full subtitle scores 0.31, `Star Wars Episode V` against its full subtitle 0.727. Both score 1.00 on coverage.

Two normalization details are load-bearing and each has a test. Punctuation becomes whitespace rather than vanishing, or `Iron-Ridge` and `Iron Ridge` share nothing. German umlauts expand before NFKD folding, or the existing `Die Fälscher` / `Die Faelscher` test breaks.

**Adoption gets its own bar.** Adopting a foreign torrent physically relocates a file the operator owns, so `UntrackedMatcher` now requires 0.9 rather than the matcher's default 0.8. Nothing is hidden: a release below that stays under Downloads, External, where `ExternalTorrents` already lists it with suggestions and one-click manual adoption.

**The tiebreak stops inflating the record.** It is now an ordering key. The confidence that is stored, filtered on, and written to the activity feed is the base score.

**Suggestions stay ungated**, deliberately. A person reads both titles before choosing, which is the same reason that path already declines the binding veto. A test asserts it.

## Recovery

The other half of the report: once a file was wrongly attached, the item page offered no way out. The only control on a row was Delete, which is the wrong answer, because the file is fine and merely filed against the wrong thing.

- **Origin.** `Library.origin_download/1` resolves the `imported_from_download_id` key that `MediaImport` already writes, and the file details modal shows which download a file came from, whether it was added outside Mydia, and the match reason. No migration: the key already existed and simply was not surfaced.
- **Not this movie / Not this show.** Detaches the row, leaves the bytes on disk, and upserts an import candidate at the file's current path so it appears in `/review`, where match search and rematch already exist. Both writes run in one transaction and emit an activity event.

## Calibration

`torrent_matcher_calibration_test.exs` is a table of invented title pairs. Every "must not match" pair was validated to score at or above 0.8 under the pre-gate matcher, so each one characterizes the defect rather than merely looking plausible. Movie, TV episode, and season-pack paths are all covered; the season-pack cases matter because that is the shape the production incident actually took.

## Known limitation

Coverage cannot reject a release whose title *extends* the library title, since every library word is present by construction. `Carrie` against `Carrie Pilby` scores 1.00. This is an accepted trade rather than an oversight: the identical shape covers a real film matched against its own full subtitle, and no rule reading only the two strings separates them. The exposure is larger for TV, which has no year term. The gate is scoped to the failure that was actually losing files, a release that does not contain the library title at all.

## Verification

Full `mix precommit` green: 10152 tests, 0 failures, plus compile, format, credo and unused-deps.

Every pre-existing matcher test that asserts a *successful* match was checked against the gate and scores 1.00 on coverage, including `Die Fälscher`, `Dr. Strangelove`, `Star Wars Episode V`, and `The Lord of the Rings`. `The Matrix` / `The Matrix Reloaded` and `Godfather` / `Godfather II` still pass the gate and are still rejected by the existing sequel penalty, unchanged.

## Follow-ups, none blocking

- The same Jaro blind spot exists in `MetadataMatcher`, `Media.title_similarity/2` and `ReleaseRanker`. Larger blast radius, worth its own issue.
- The recovery action is on version rows but not `episode_file_row/1`, so a single misattached episode has no in-app route back to review yet.
- `detach_to_review/3` is private, so its atomicity test re-composes the two writes rather than calling it; extracting the transaction body would let the test guard the wiring rather than only Ecto's semantics.

## Summary by CodeRabbit

* **New Features**
  * Return incorrectly matched media files to the review inbox without deleting them from disk.
  * View a file’s originating download and matching details in the file information panel.
  * Record and display “Returned to review” activity events.

* **Bug Fixes**
  * Improved torrent matching to reject insufficient title matches and preserve accurate confidence scores.
  * Raised the confidence requirement for automatically adopting untracked downloads.
  * Added safer handling for invalid or legacy files returned to review.


arsfeld:
@coderabbitai review

arsfeld:
@coderabbitai full review

arsfeld:
Good catch on the ownership check, fixed in 82d113796.

`not_this_item/2` now gates on `owns_media_file?/2` before fetching anything, which is the same guard `reclassify/3` already uses a few functions down in that file for exactly this reason. Its comment there spells out the risk, and I had simply not applied it to the new handler.

Added a regression test that submits another item's file id and asserts the foreign row is untouched and never reaches the review inbox. I confirmed it discriminates by bypassing the guard and watching it fail, rather than assuming it would.

PR #675, opened 2026-09-02
fix(indexers): stop a magnet redirect crashing the grab

arsfeld:
## The bug

A monitored show searched, ranked 100 results, selected a season pack, and then never downloaded it. No download row was created, no error surfaced, and the application log showed nothing at all.

The cause, recovered from `oban_jobs.errors` on a production instance:

```
** (ArgumentError) invalid scheme "magnet" for url: magnet:?xt=urn:btih:...
    (finch 0.23.0) lib/finch/request.ex:152: Finch.Request.parse_url/1
    (req 0.7.1) lib/req/steps.ex:1482: Req.Steps.redirect/1
```

`CardigannDownload.resolve/3` fetched the download URL through `execute_http_request/5`, which passes `redirect: definition.follow_redirect`. That defaults to `true` (`cardigann_parser.ex:180`), and the indexer had no stored cookies, so the `attach_cookies/4` branch that forces `redirect: false` never ran. The tracker answered `302 Location: magnet:?xt=...`, Req followed it, and Finch rejected the scheme.

`execute_http_request/5` converts returned error tuples only, so the raise passed straight through `initiate_season_pack_download/5` without reaching either its success or its failure branch. It killed the Oban worker, retried identically twice, and discarded. On the instance where this was found, all 6 failed `TVShowSearch` jobs carried this same error.

Both cookie states were broken, differently:

- **No cookies:** `redirect: true`, Req follows and raises. This is the observed crash.
- **Cookies present:** `redirect: false` already, so the 302 returns an empty body, every selector matches nothing, and the grab fails with "no download selector matched the response".

## What changed

**`execute_download_request/6`** (`cardigann_search_engine.ex`) resolves redirects one hop at a time with `redirect: false` and short-circuits the moment a `Location` names a magnet, without issuing a second request. Each hop rebuilds its options through the existing `build_request_options/5`, so `attach_cookies/4` re-runs `cookie_disposition/2` against the new URL. That is the per-hop origin check 126cdead9 deliberately deferred: a session cookie is re-authorised for every hop rather than carried across one blindly. A magnet `Location` resolves with no further request, so no credential reaches the host that named it.

**`CardigannDownload`** consumes the new tagged shapes and short-circuits on a magnet in both the infohash and selector arms. `resolve/3`'s public contract is unchanged, so no caller outside the file needed an edit.

**`safe_request/3`** converts a raised Req/Finch error into the `{:error, %Error{}}` the Indexers contract already promises, and both request paths now use it. An indexer controls both the URL it advertises and the `Location` it redirects to, so a raise there is remote input reaching the top of an Oban worker.

**`Mydia.Jobs.ErrorLogger`** logs Oban job failures. `Mydia.Jobs.Broadcaster` already listened for `[:oban, :job, :exception]`, but only to broadcast status over PubSub and record an audit event. Nothing wrote the failure to the application log and `Oban.Telemetry.attach_default_logger/1` was never called, which is why three crashed attempts and a permanent discard were indistinguishable from a job that had nothing to do.

## Why not reuse the existing redirect follower

`Mydia.Downloads.Queue` has one (`follow_to_final_url/3`, `queue.ex:1573`) that also terminates on `magnet:`. It is not shared here on purpose. It probes with HEAD and returns a final URL for the caller to re-fetch, and it carries one flat cookie header across every hop. The Cardigann path needs the body from the final hop and needs its cookie decision re-evaluated per hop against the definition's trusted origins. Reusing it as-is would have reintroduced the cross-host cookie leak that 126cdead9 closed. Happy to revisit if a reviewer prefers a shared resolver taking an opts-builder closure.

## Testing

Full suite green. Format and Credo clean. Eight tests added:

- 302 to a magnet resolves to the magnet without a second request
- an ordinary http redirect is still followed to the final body
- a redirect loop gives up rather than recursing forever
- a 3xx with no `Location` header errors rather than crashing
- a redirect to an unsupported scheme returns an error instead of raising
- end to end through `resolve/3`: magnet redirect, landing-page redirect, and a session not carried onto a cross-host redirect
- `ErrorLogger` puts the worker, attempt and failure reason in the log message

Two false-greens were caught and fixed during review, both of which passed while proving nothing:

- The first `ErrorLogger` test's `"invalid scheme"` assertion was satisfied by `CrashReporter.TowerReporter`'s telemetry handler dumping raw metadata into a separate warning line, not by this module. The test now calls `handle_event/4` directly so only this module's output can satisfy it.
- The handler originally passed the failure reason as `oban_error:` metadata, but this app's Logger formatter whitelists metadata keys and no `oban_*` key is registered in `config.exs`, `dev.exs` or `test.exs`. A real operator line would have read `Oban job failed: <worker> (attempt 3/3)` with the exception invisible. `ExUnit.CaptureLog` does not apply that whitelist, so the test was green. The reason is now interpolated into the message string, and the replacement test was verified to fail when it is moved back to metadata.

## Known follow-ups, deliberately not in this PR

- `run_req/3` uses a bare `rescue exception ->`, broader than the single documented failure mode. It now covers every Cardigann **search** request, not just downloads, so a future malformed-options bug would surface as a generic "search failed" rather than crashing loudly. It fails safe (logged, reason preserved in the returned tuple) and `exit`/`throw` are unaffected. Narrowing it to `rescue e in ArgumentError` and re-raising the rest is the obvious improvement.
- No test covers the infohash arm's magnet short-circuit (`cardigann_download.ex:150`); all three end-to-end tests drive the selectors arm. Verified correct by inspection, structurally symmetric to the tested arm.
- Pre-existing and untouched here: when `[:oban, :job, :exception]` fires outside a sandboxed process in the test env, `Broadcaster`'s handler crashes on a missing Ecto sandbox checkout and `:telemetry` permanently auto-detaches it for the rest of that suite run. Worth its own issue.
- `execute_http_request/5` and `execute_download_request/6` still carry near-identical `Req.TransportError` conversion clauses.
- `apply_rate_limit/1` runs per redirect hop, so a definition with `request_delay` pays it up to 5 times on a maxed chain. Defensible, since each hop is a real outbound request to a rate-limited host.

## Not fixed, and why

`TorrentHash.download_torrent/1` (`torrent_hash.ex:294`) and `Blackhole` (`blackhole.ex:241`) both call `Req.get/2` with no redirect guard and carry the same crash. Both are reachable only through `TorrentHash.extract({:url, url})`, and `Queue.prepare_torrent_input/2` produces a `{:url, _}` only from its "Unknown format" fallback, which no configured indexer currently hits. Fixing them now would be speculative work on an unreachable path. If a magnet-scheme `ArgumentError` ever appears in `oban_jobs.errors` with a `TorrentHash` or `Blackhole` frame, that is the trigger.

## Summary by CodeRabbit

- **New Features**
  - Download links can follow redirects, resolve directly to magnet links, and continue processing redirected landing pages.
  - Cross-site redirects no longer carry session cookies.
- **Bug Fixes**
  - Download and search requests now handle network errors, redirect loops, missing destinations, unsupported redirect schemes, and disabled redirect following without crashing.
  - Failed background jobs are now clearly recorded in application logs with useful failure details.

### DIFF
--- a/lib/mydia/application.ex
+++ b/lib/mydia/application.ex
@@ -26,6 +26,18 @@ defmodule Mydia.Application do
 
     children = children()
 
+    # Attached before the supervisor starts, unlike its sibling
+    # Mydia.Jobs.Broadcaster below: Oban is one of the children
+    # Supervisor.start_link/2 is about to start, and it can begin executing
+    # (and failing) jobs immediately. Attaching after start_link/2 leaves a
+    # window where a job can fail before anything is listening, and the
+    # failure goes unlogged, exactly what this handler exists to prevent. This
+    # is safe to do this early because the handler only calls Logger.error and
+    # touches no Repo. Broadcaster cannot move here: its handler calls
+    # Mydia.Jobs.list_executing_jobs(), which needs the Repo child already
+    # running.
+    Mydia.Jobs.ErrorLogger.attach()
+
     # See https://hexdocs.pm/elixir/Supervisor.html
     # for other strategies and supported options
     opts = [strategy: :one_for_one, name: Mydia.Supervisor]
@@ -38,10 +50,6 @@ defmodule Mydia.Application do
       reset_stale_jobs()
       # Attach Oban job broadcaster for real-time job status updates
       Mydia.Jobs.Broadcaster.attach()
-      # Log Oban job failures. Without this a raising worker is silent: the
-      # crash lands only in oban_jobs.errors and a discarded job looks like a
-      # job that simply had nothing to do.
-      Mydia.Jobs.ErrorLogger.attach()
       # Register download client adapters after supervisor has started
       Mydia.Downloads.register_clients()
       # Register indexer adapters after supervisor has started
--- a/lib/mydia/application.ex
+++ b/lib/mydia/application.ex
@@ -38,6 +38,10 @@ defmodule Mydia.Application do
       reset_stale_jobs()
       # Attach Oban job broadcaster for real-time job status updates
       Mydia.Jobs.Broadcaster.attach()
+      # Log Oban job failures. Without this a raising worker is silent: the
+      # crash lands only in oban_jobs.errors and a discarded job looks like a
+      # job that simply had nothing to do.
+      Mydia.Jobs.ErrorLogger.attach()
       # Register download client adapters after supervisor has started
       Mydia.Downloads.register_clients()
       # Register indexer adapters after supervisor has started
--- a/lib/mydia/indexers/cardigann_download.ex
+++ b/lib/mydia/indexers/cardigann_download.ex
@@ -144,9 +144,17 @@ defmodule Mydia.Indexers.CardigannDownload do
          variables,
          user_config
        ) do
-    with {:ok, response} <-
-           infohash_source(definition, infohash, download_url, before_response, user_config),
-         {:ok, hash} <- match_selector(response, Map.get(infohash, :hash), variables, "hash"),
+    with {:ok, source} <-
+           infohash_source(definition, infohash, download_url, before_response, user_config) do
+      case source do
+        {:magnet, magnet} -> {:ok, {:magnet, magnet}}
+        {:response, response} -> build_magnet_from_response(response, infohash, variables)
+      end
+    end
+  end
+
+  defp build_magnet_from_response(response, infohash, variables) do
+    with {:ok, hash} <- match_selector(response, Map.get(infohash, :hash), variables, "hash"),
          {:ok, title} <- match_title(response, Map.get(infohash, :title), variables) do
       case TorrentHash.build_magnet(hash, title) do
         nil ->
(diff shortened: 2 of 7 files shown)


arsfeld:
All three review findings were verified against the code and fixed in 316448728.

**1. `follow_download_redirect/7` ignored `definition.follow_redirect`.** Correct, and the most consequential of the three. After the redirect rework, `follow_redirect` was referenced only in `build_request_options/5`, whose value this path then overrode with `redirect: false`, so a definition setting `followredirect: false` had its download redirects followed anyway. Every test on this branch used the default `true`, so nothing caught it.

Fixed with the distinction that matters: a `magnet:` Location still resolves regardless of the setting, because reading a header issues no request and leaks no credential, and the magnet is the payload being handed over. An http(s) Location now only gets followed when `follow_redirect` is true; when false it returns an error naming the status and Location, matching `validate_response/1`'s existing wording for the same case on the search path. Two regression tests cover both.

**2. `ErrorLogger.attach/0` ran after Oban had started.** Correct. It sat inside the `with {:ok, pid} <- Supervisor.start_link(...)` block and Oban is one of those children, so a job failing in that window went unlogged, which is exactly what the module exists to prevent. Moved before `start_link/2`. Safe specifically for this handler because it only calls `Logger.error` and touches no Repo. `Broadcaster.attach()` deliberately stays put, since its handler calls `Mydia.Jobs.list_executing_jobs()` and needs the Repo running.

**3. The test left the application's handler globally detached.** Correct, and independently flagged during this branch's own review. `setup` now attaches and records whether that call was the one that succeeded, and `on_exit` detaches only in that case, so a boot-time attach is left alone.

One suggestion not taken: the startup test for finding 2. Asserting application boot ordering would need the supervision tree stood up around a pre-seeded failing job, which is brittle and disproportionate to a one-line move whose correctness is visible at the call site. Happy to add it if you disagree.

PR #681, opened 2026-09-03
fix(p2p): stop the connection dropping and reconnecting for no reason

arsfeld:
Three fixes for one reported symptom: a player connected to the production instance shows its connection dropping and reconnecting, and keeps working afterward.

Two independent defects produce that, and they are fixed separately. A third change removes a resource-exhaustion vector the second one would otherwise have introduced.

## 1. A stale connection closing reported a live peer as gone

A peer routinely holds several connections at once. Production has caught the player opening eight inside 100ms, and `connected_peers` is keyed by `peer_id`, so each new connection replaces the previous entry while the replaced connection's `handle_connection` task keeps running.

When one of those stale connections finally timed out, it announced `Event::Disconnected` for a peer that was still connected and still serving requests on a newer connection. The `stable_id` guard added for T-809 in 865a9ebf3 kept the map honest, refusing to evict the live entry, but the event went out regardless and every consumer downstream believed it: the Elixir host drops the peer from `state.connected_peers`, and a player redials a link that never broke. The journal shows it plainly, six `Connection closed ... timed out` plus `Peer Disconnected` lines in a row for a single peer.

`handle_connection` cannot make this call. It knows only that its own connection ended, never whether a newer one has since taken over the `peer_id`. Only the event loop owns `connected_peers`, so the announcement moved into `prune_disconnected_peer`, under the same `stable_id` guard that already decides whether the entry is really gone.

This is the better match for the reported symptom: it produces a drop-and-reconnect with no crash and no restart.

## 2. A crashing resolver took the whole host down

`Mydia.P2p.Server` resolved peer requests inside its own process with no exception handling, so an exception escaping a resolver did not fail one query, it killed the P2P host. The supervisor rebuilt the endpoint and every paired player had to redial. Three such restarts in seven days, all with `Last message: request_received graphql`:

| When | Exception | Operation |
| --- | --- | --- |
| Aug 30 00:25 | `ArgumentError: unknown registry: MydiaWeb.Endpoint.Registry` via `Absinthe.Subscription.publish` | `updateEpisodeProgress` |
| Sep 01 22:37 | `Exqlite.Error: Database busy` via `ForeignKeyGuard.run/3` and `Playback.save_progress/4` | `updateEpisodeProgress` |
| Sep 02 19:35 | the registry error again | `updateMovieProgress` |

The third fired during a SIGTERM redeploy, so the endpoint was already stopped and that one is a shutdown race rather than a fault of its own.

GraphQL, pairing and `read_media` now run off the GenServer through a shared `serve_request/4`, which is what `stream_hls_response` has always done for throughput ("so we don't block the GenServer"). Beyond containing the crash, this also stops one slow query from delaying every other peer's request and the host's own `peer_connected` events, which is the same lack of isolation showing up as latency instead of a restart.

`serve_request/4` answers exactly once, including when the handler raises. A peer that gets nothing back waits out the Rust core's 30s `RESPONSE_TIMEOUT` instead, which is a worse failure than an error it can act on immediately.

Pairing was checked before being parallelized rather than assumed safe: `ClaimRateLimiter` uses an atomic `:ets.update_counter` with an `insert_new` retry and is already reachable concurrently from the HTTP path, so the brute-force guess limit does not depend on the GenServer serializing it.

## 3. Bounding what that spawns

Plain `Task.start` per request would have been a regression introduced by fix 2. `accept_inbound` gates on ALPN alone, so a peer needs no credential to make the host spawn these, and this subsystem already had two memory-exhaustion holes closed in 865a9ebf3. The tasks run under a bounded `Task.Supervisor` (`max_children: 256`); past the limit the peer is refused with "Server busy" rather than the host accumulating work it will never finish.

## Tests

`a_stale_connection_closing_does_not_report_a_live_peer_as_gone` gives two clients the same secret key so both connections land under one `peer_id` on the server, drops the first, and asserts nothing is announced while the second is live. It then drops the second and asserts the disconnect does arrive, so the first assertion cannot pass on a dead event path. Without the fix it fails with exactly one spurious disconnect.

`Mydia.P2p.ServerGraphQLTest` resolves against a purpose-built `CrashingSchema` rather than a production resolver. Two candidates for a "naturally raising" query did not hold up: `movie(id:)` is answered by the authentication middleware before the resolver runs, and once authenticated a malformed id simply misses on SQLite and returns a rescued "Movie not found". What is worth pinning is the isolation contract, not whichever resolver happens to lack a `rescue` this month. Verified in both directions: with the `rescue`/`catch` removed, three of the five fail with the exception and the exit propagating out of the caller.

It also pins the remote-access gate for p2p GraphQL, which had no coverage at all and which fix 2 moved.

## Not fixed here

The player opens eight connections to one peer in 100ms and fires eleven duplicate `Devices` queries, which suggests `ensureConnected` is not deduped across concurrent GraphQL calls. Fix 1 makes the server stop lying about the resulting disconnects, but the client-side duplication is its own defect and is untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented failed GraphQL resolvers and remote requests from crashing remote access or disconnecting other peers.
  * GraphQL failures now return consistent error responses while subsequent requests continue normally.
  * Improved handling of multiple connections to the same peer, preventing stale connections from triggering incorrect disconnection notifications.
  * Remote GraphQL requests remain blocked when remote access is disabled.
  * Hung requests now time out instead of remaining active indefinitely.

* **Reliability**
  * Limited concurrent remote requests to help maintain service responsiveness during high demand.
  * Requests exceeding capacity now receive a clear “Server busy, try again” response.

### DIFF
--- a/lib/mydia/application.ex
+++ b/lib/mydia/application.ex
@@ -230,6 +230,12 @@ defmodule Mydia.Application do
         # Must precede P2p.Server: it creates the instance identity the p2p node
         # and every pairing depend on, and seeds RemoteAccess.enabled?/0.
         Mydia.RemoteAccess.Provision,
+        # Serves peer requests off Mydia.P2p.Server, one task each. Bounded
+        # because iroh gates inbound connections on ALPN alone, so a peer needs
+        # no credential to make the host spawn these, and `max_children` is what
+        # keeps that from becoming the third memory-exhaustion hole in this
+        # subsystem. Over the limit the request is refused rather than queued.
+        {Task.Supervisor, name: Mydia.P2p.RequestSupervisor, max_children: 256},
         Mydia.P2p.Server,
         # Resume active pairing claims on startup
         Mydia.RemoteAccess.ResumeClaims
--- a/lib/mydia/application.ex
+++ b/lib/mydia/application.ex
@@ -230,12 +230,18 @@ defmodule Mydia.Application do
         # Must precede P2p.Server: it creates the instance identity the p2p node
         # and every pairing depend on, and seeds RemoteAccess.enabled?/0.
         Mydia.RemoteAccess.Provision,
-        # Serves peer requests off Mydia.P2p.Server, one task each. Bounded
-        # because iroh gates inbound connections on ALPN alone, so a peer needs
-        # no credential to make the host spawn these, and `max_children` is what
-        # keeps that from becoming the third memory-exhaustion hole in this
-        # subsystem. Over the limit the request is refused rather than queued.
-        {Task.Supervisor, name: Mydia.P2p.RequestSupervisor, max_children: 256},
+        # Serves peer requests off Mydia.P2p.Server. Bounded because iroh gates
+        # inbound connections on ALPN alone, so a peer needs no credential to
+        # make the host spawn these, and `max_children` is what keeps that from
+        # becoming the third memory-exhaustion hole in this subsystem. Over the
+        # limit the request is refused rather than queued.
+        #
+        # Two children per in-flight request, so this caps concurrency at 256:
+        # one runs the handler, the other waits on it with a deadline and kills
+        # it if it hangs. A handler blocked on an unresponsive filesystem
+        # returns on its own schedule or never, and a bound whose slots never
+        # come back just fails every later request instead.
+        {Task.Supervisor, name: Mydia.P2p.RequestSupervisor, max_children: 512},
         Mydia.P2p.Server,
         # Resume active pairing claims on startup
         Mydia.RemoteAccess.ResumeClaims
--- a/lib/mydia/p2p/server.ex
+++ b/lib/mydia/p2p/server.ex
@@ -249,22 +249,23 @@ defmodule Mydia.P2p.Server do
   def handle_info({:ok, "request_received", "pairing", request_id, req}, state) do
     Logger.info("P2P Request: Pairing from #{req.device_name}")
 
-    response =
-      if RemoteAccess.enabled?() do
-        handle_pairing_request(req)
-      else
-        Logger.info("P2P Request: Pairing refused, remote access is disabled")
+    serve_request(state.resource, request_id, "pairing", fn ->
+      response =
+        if RemoteAccess.enabled?() do
+          handle_pairing_request(req)
+        else
+          Logger.info("P2P Request: Pairing refused, remote access is disabled")
 
-        %P2p.PairingResponse{
-          success: false,
-          error: "Remote access is disabled on this server"
-        }
-      end
+          %P2p.PairingResponse{
+            success: false,
+            error: "Remote access is disabled on this server"
+          }
+        end
 
-    # Wrap in tagged enum tuple as expected by NIF
-    response_enum = {:pairing, response}
+      # Wrap in tagged enum tuple as expected by NIF
+      {:pairing, response}
+    end)
 
-    P2p.send_response(state.resource, request_id, response_enum)
     {:noreply, state}
   end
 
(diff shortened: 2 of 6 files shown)


arsfeld:
Applied in e90be46fb. Both `catch` branches now log via `Exception.format(kind, reason, __STACKTRACE__)`, matching the `rescue` branch above each of them. `serve_request/4` had the identical shape and was fixed alongside `run_graphql/5`, which the review had not seen yet. The error returned to the peer is unchanged.

arsfeld:
@coderabbitai full review

arsfeld:
All four findings addressed.

**The guard covers only one of the two orderings (Major).** Correct, and the important one. I had handled the displaced connection closing last and missed the current connection closing first, which removed the entry and announced the peer gone while it was still serving requests on the older connection. The doc comment claimed the event meant "last live connection" and the map could not prove it.

`connected_peers` is now `HashMap<String, HashMap<ConnectionId, Connection>>`, so `Event::Disconnected` fires only when a peer's last connection closes. Reads that want a single connection take the most recent. `DebugCloseConnection` closes all of a peer's connections, so "close this peer" means the peer really is gone. Fixed in de80f2e8e.

**Do not assume `stable_id` is unique across connection lifetimes (Minor).** Agreed, and it folds into the same change: connections are keyed by a monotonic counter now. Worth noting this exposure predates the branch, it came in with the T-809 guard in 865a9ebf3.

**Wait for the server's second `Event::Connected` (Minor).** Right that the test was racing, and right about why: `connected_peers == 1` is satisfied by one connection, so the server might not have accepted the second yet and the silence proved nothing. It waits on a new `debug_connection_count` hook instead, which is also the only way to distinguish a peer connected once from the same peer connected twice. A second test covers the ordering the old guard missed.

Both tests were verified against simulated single-slot semantics and fail there: they cannot even reach the two-connection state the bug needs.

**Return stable errors to peers after caught failures (Major, CWE-209).** Agreed. Peers now get a flat `"Request failed"` and the full `Exception.format/3` output including the stacktrace goes to the log. The `{:error, reason}` branch also ran binaries through `inspect/1`, so the player would have rendered the quotes verbatim; a reason that is already a sentence is sent as one. Fixed in d930fa805.

The tests are stronger for it: rather than asserting the resolver's message reached the caller, they now assert the peer is told nothing and the operator is told everything.

Local verification: `mix precommit` green, 10594 tests 0 failures; Rust 38 tests, `cargo fmt` clean, no new clippy warnings.

arsfeld:
@coderabbitai full review

arsfeld:
**Add a per-request timeout that terminates the worker task (Major).** Agreed, and it is a failure mode this PR introduced: bounding the pool without a way to reclaim slots just trades an unbounded spawn for permanent exhaustion. You are also right that the Rust `RESPONSE_TIMEOUT` is not a substitute; it abandons the peer's side of the exchange and cannot stop a BEAM process.

Each request now runs in a worker that the outer task waits on with a 30s deadline, matching the transport's, and kills with `Task.shutdown(:brutal_kill)` when it expires. The peer gets `@opaque_failure`. `max_children` moved to 512 because a request now occupies two children, so the effective cap on concurrent requests is unchanged at 256.

Covered by `Mydia.P2p.ServerRequestTimeoutTest`, verified in both directions: with the deadline removed the hang test fails because the slot is never returned.

**Restrict `read_media` to authorized server-owned files (Major, CWE-22).** Confirmed, and thank you for it. `req.file_path` is peer-controlled, only checked for existence, and `accept_inbound` admits on ALPN alone, so no credential is needed.

Deliberately **not** fixed in this PR, for three reasons:

1. It is pre-existing rather than a regression here. The handler already carried the comment `SECURITY: In production, verify path is within allowed directories!`; this PR only moved it into a task.
2. Nothing sends the request. `ReadMedia` is defined in the Rust `MydiaRequest` enum, but the player has no references to it and the core never constructs one, so the likely fix is to refuse it outright rather than build media-id resolution. That is a protocol decision.
3. This PR is about connection stability and crash isolation. Folding an auth change into it makes both harder to review.

It is being tracked separately and will be fixed on its own. Flagging it clearly here so it is not mistaken for something this PR addressed.

arsfeld:
@coderabbitai review

PR #683, opened 2026-09-03
fix(media): size media file rows from their column, not the window

arsfeld:
The Media Files card was fixed for phones in #616 by stacking its action strip below `sm:`. This does the same for the TV episode file row, and in the process replaces the `sm:` breakpoint on both surfaces, because measuring the page showed `sm:` is the wrong signal here.

## Why `sm:` was wrong

Two layout switches sit between the browser window and these columns: the app drawer becomes permanent at `lg` (256px), and the show page's own left rail appears at `md` and widens to 20rem at `lg`. Both take width away at the moment a viewport breakpoint grants it, so a wide window does not mean a wide column. The episode column is 537px at a 640px viewport and 274px at 1024px.

Filename column widths measured on the running app, with a 92-character episode basename and an 84-character show-level one:

| viewport | episode | movie |
| --- | --- | --- |
| 375 | 69px (8 chars) | 287px |
| 640 | 326px | 280px |
| 768 | 158px (19 chars) | 96px (11 chars) |
| 1024 | 78px (9 chars) | **16px (2 chars)** |
| 1440 | 494px | 432px |

1024px is the worst case on both, not the phone. The Media Files card, already "fixed", rendered two characters of an 84-character filename on a 1024px laptop.

Separately, the collapsed episode row's toolbar rendered 50px past the right edge of a 1024px viewport, scrolling the document sideways with every episode closed.

## What changed

Two named Tailwind container queries replace `sm:`: `eprow` on the per-episode wrapper, `mfrow` on the Media Files card body. Each row now responds to the width of the column it is in.

The container is named rather than bare because `show.html.heex` already declares an unnamed `@container` around the whole page grid; a bare `@lg:` would silently query that instead, and it measures 744px where these columns measure under 300px.

Below the threshold the row stacks, the action strip takes its own line, and its buttons grow from 24px to 40px. Above it, everything renders as it does today.

The threshold is `@md` (28rem / 448px), against the container's **content box**. That distinction is the subtle part and is now recorded in a comment at both declarations: an earlier `@lg` (512px) left the 900px case clearing by 3px, where a vertical scrollbar appearing was enough to flip the layout.

## Result

| viewport | episode | movie | stacked |
| --- | --- | --- | --- |
| 375 | 245px | 287px | yes |
| 414 | 284px | 326px | yes |
| 640 | 311px | 265px | no |
| 768 | 319px | 345px | yes |
| 900 | 275px | 213px | no |
| 1024 | 224px | 250px | yes |
| 1280 | 304px | 242px | no |
| 1440 | 464px | 402px | no |

Behaviour at 640, 900, 1280 and 1440 is unchanged. Horizontal document scroll at 1024px is gone.

## Testing

- `test/mydia_web/features/media_file_row_width_test.exs` measures rendered widths in a real browser across all eight viewports and asserts a 200px readable floor, plus a document-overflow check at 1024px. Each assertion was proved able to fail by reinstating `sm:`, since a geometry probe whose selector misses passes silently.
- Class-level assertions in the three LiveView test files use a regex helper so that `@md/eprow:btn-xs` does not satisfy a check for a bare `btn-xs`.
- `mix precommit`: 10,599 tests, 0 failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved media file and episode row layouts in narrow columns, including when side panels reduce available space.
  * Prevented long filenames and row actions from causing horizontal page overflow.
  * Preserved readable filename widths across common screen sizes.
  * Improved responsive behavior for episode details, status information, extras, versions, and action buttons.

* **Tests**
  * Added coverage for responsive layouts, readable filenames, tap targets, and overflow prevention across multiple viewport sizes.

### DIFF
--- a/lib/mydia_web/live/media_live/show/components.ex
+++ b/lib/mydia_web/live/media_live/show/components.ex
@@ -969,8 +969,11 @@ defmodule MydiaWeb.MediaLive.Show.Components do
                   width. --%>
             <ul class="menu w-full flex-nowrap bg-base-100 rounded-box p-0 mt-2">
               <li :for={file <- @extras} id={"extra-#{file.id}"} class="min-w-0 flex-nowrap">
-                <div class="min-w-0 items-stretch flex flex-col gap-2 p-4 rounded-none sm:flex-row sm:items-center sm:justify-between sm:gap-4">
-                  <div class="min-w-0 sm:flex-1 flex flex-col gap-1">
+                <div
+                  id={"extra-row-#{file.id}"}
+                  class="min-w-0 items-stretch flex flex-col gap-2 p-4 rounded-none @lg/mfrow:flex-row @lg/mfrow:items-center @lg/mfrow:justify-between @lg/mfrow:gap-4"
+                >
+                  <div class="min-w-0 @lg/mfrow:flex-1 flex flex-col gap-1">
                     <p
                       id={"extra-name-#{file.id}"}
                       class="text-sm font-mono truncate"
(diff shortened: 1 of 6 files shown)


arsfeld:
@coderabbitai review

PR #712, opened 2026-09-05
fix(media): scope provider-id uniqueness by media type, close the check-then-write race

### LINKED ISSUE #473: Scope provider-id uniqueness by media type, and close the check-then-write race
Two findings deferred from #456, both on the provider-id write path in `Mydia.Media.ExternalIds`.

## 1. `tmdb_id` and `tvdb_id` uniqueness is global, not scoped by type

TMDB numbers movies and TV series in independent namespaces, but `media_items` carries a single global unique index on `tmdb_id` (and another on `tvdb_id`). A movie and a TV show that legitimately share a numeric id are treated as conflicting rows.

Two visible symptoms:

- Adding a TV show whose `tmdb_id` equals an existing movie's still shows the raw `Failed to add: tmdb_id: has already been taken`.
- `ExternalIds.lookup/2` does not filter by `type`, so a cross-referenced id can be reported as owned by a row of the other media type.

The fix is composite unique indexes on `(type, tmdb_id)` and `(type, tvdb_id)`, plus a `type` filter on the provider-id lookups. It needs a migration that resolves whatever collisions an existing library already holds, and it has to work on both adapters: Postgres can alter the indexes directly, SQLite needs the table-rebuild path in `Mydia.Repo.Migrations.Helpers`.

`test/mydia/media/add_test.exs` has a test pinning the current movie-vs-TV behavior. It is the baseline to change.

## 2. Conflict detection and the provider-id write are not atomic

`ExternalIds.put_free_ids/3` reads ownership and returns attrs; the caller performs the insert or update afterwards. Two concurrent operations can both observe an id as free, and the loser fails on the unique index with a constraint error instead of the controlled duplicate-provider warning flow.

Closing this means either moving the write into `ExternalIds` or wrapping the call sites in a transaction that catches the constraint error, re-reads the owner via `conflicting_item/3`, emits `media_item.duplicate_provider_id`, and returns the same controlled result.

Low priority on its own: the window needs two writers learning the same third-party id in the same instant on a single-node instance, and the failure mode is the pre-#456 behavior. Worth doing together with the index change since both touch the same write path.

Original review threads: https://github.com/getmydia/mydia/pull/456#discussion_r3788736112 and https://github.com/getmydia/mydia/pull/456#discussion_r3788736115

arsfeld:
Closes #473.

TMDB and TVDB number movies and series in independent namespaces, but `media_items` carried a single global unique index on `tmdb_id` and another on `tvdb_id`. A movie and a show that legitimately share a number were treated as conflicting rows, so adding such a show surfaced the raw `tmdb_id: has already been taken`, and provider-id lookups reported a row of the wrong media type as the owner of an id.

Separately, `ExternalIds.put_free_ids/3` read ownership and the caller wrote afterwards, so two writers could both see an id as free and the loser failed on the unique index instead of getting the controlled duplicate-provider warning.

## What changed

**Indexes.** The two global unique indexes become partial composite ones on `(type, tmdb_id)` and `(type, tvdb_id)`. This is a pure relaxation: the composite index is strictly weaker than the global one, so no existing row can fail it and `up` needs no data step. `down` refuses with the offending provider ids and titles named rather than letting the adapter raise a bare unique violation partway through a rollback, and both guards run before any DDL so a refusal leaves the schema untouched.

Both indexes stay partial deliberately. SQLite reports a partial unique index violation using the literal index name and a plain one using a column list that `ecto_sqlite3` rewrites into an Ecto-convention name. Keeping them partial makes both adapters report the same name, which is what `MediaItem.changeset/2`'s `unique_constraint` calls have to match. Those calls are renamed accordingly; without that a same-type collision would raise `Ecto.ConstraintError` instead of returning a changeset.

**Lookups.** `ExternalIds` now requires an explicit `:type` and routes ownership checks through the type-validating `Media.find_by_external_ids/2`. The untyped `Media.get_media_item_by_tmdb/2` and `get_media_item_by_tvdb/2` are deleted and all callers moved over, so one typed lookup remains in `Media` instead of three overlapping ones and an untyped call is no longer spelled the same way as a typed one. Three production callers were enforcing the old global rule in application code: `MediaRequests.check_duplicate_media/1` rejected a TV request whose `tmdb_id` matched a movie before any insert was attempted, and the scanner and Discover existence probes reported the wrong row as owner.

**The race.** `ExternalIds.write/3` catches a provider-id constraint error, re-reads every provider id in the attrs, reports and drops the ones still taken, and retries once. Dropping all of them in a single pass is what makes one retry always sufficient: a database aborts at the first violated index and never reports a second from the same attempt, so dropping only the id it named could still collide on the retry.

No transaction is used to close the race, because none would. At PostgreSQL's default `READ COMMITTED` both writers still read the id as free and one still fails at write time, and SQLite's deferred transactions behave the same. The catch-and-re-read does the work alone, and it also covers stale reads seconds old rather than only the instant.

Each attempt does run in its own savepoint. On PostgreSQL a statement that fails a constraint check aborts the whole ambient transaction (SQLSTATE `25P02`), and two of the four wrapped call sites run inside a transaction the caller opened, so the re-read that follows a collision raised instead of degrading. Rolling back to the savepoint rather than merely releasing it is the load-bearing detail: releasing a savepoint whose statement already failed does not clear the aborted state. `Repo.transaction(fun, mode: :savepoint)` cannot express this, because for a nested transaction on the same connection `DBConnection.transaction/3` runs the function directly and discards its opts, so no `SAVEPOINT` is ever sent.

## Testing

Full suite green at 10760 tests. The migration, changeset, `ExternalIds` and `write/3` behavior are covered on both adapters, and the constraint-name agreement between SQLite and PostgreSQL is asserted directly, since that is the piece the default SQLite-only suite cannot see.

The race regression test drives a genuine same-type collision through `approve_request/3`, using a telemetry handler to land the colliding row between the pre-flight read and the insert on the same connection inside the same transaction. Verified on PostgreSQL that it fails with `ERROR 25P02` without the savepoint and passes with it.

## Follow-ups, not in this PR

- Five call sites write `tmdb_id`/`tvdb_id` without going through `ExternalIds` (`import_lists.ex`, `jobs/import_list_auto_add.ex`, `media/provider_switch.ex`, `controllers/api/media_controller.ex`, `live/search_live/index.ex`). Not regressions, since the changeset-level constraint rename protects them from the crash class, but they surface an ordinary changeset error rather than the graceful drop-and-warn.
- The migration `down` guard's raw SQL is untested on PostgreSQL: `Mydia.MigrationTestRepo` is SQLite-only and the PostgreSQL job runs `up` via `ecto.migrate`, never `down`.
- No test covers the window where a provider id's owner vanishes between the failed write and `drop_taken/3`'s live re-read.

## Summary by CodeRabbit

* **Improvements**
  * Media provider IDs are now scoped by media type, allowing movies and TV shows to share the same TMDB or TVDB ID.
  * Duplicate detection distinguishes same-type conflicts from valid cross-type matches.
  * Media creation and updates handle provider-ID conflicts more reliably, including race conditions.
* **Bug Fixes**
  * Invalid media types now produce validation errors without triggering lookup failures.
  * Provider-ID uniqueness conflicts return clear validation feedback instead of unexpected errors.

### DIFF
--- a/lib/mydia/library/metadata_enricher.ex
+++ b/lib/mydia/library/metadata_enricher.ex
@@ -131,10 +131,10 @@ defmodule Mydia.Library.MetadataEnricher do
          match_result,
          config
        ) do
+    provider_key = if provider_type == :tvdb, do: :tvdb, else: :tmdb
+
     existing_item =
-      if provider_type == :tvdb,
-        do: Media.get_media_item_by_tvdb(id),
-        else: Media.get_media_item_by_tmdb(id)
+      Media.find_by_external_ids(%{provider_key => id}, type: media_type_to_string(media_type))
 
     case existing_item do
       nil ->
--- a/lib/mydia/library/metadata_enricher.ex
+++ b/lib/mydia/library/metadata_enricher.ex
@@ -452,19 +452,45 @@ defmodule Mydia.Library.MetadataEnricher do
     end
   end
 
-  defp persist_media_item(%MetadataPreparation{operation: :create, media_item_attrs: attrs}) do
-    Media.create_media_item(attrs, skip_episode_refresh: true)
+  defp persist_media_item(%MetadataPreparation{
+         operation: :create,
+         media_type: media_type,
+         media_item_attrs: attrs
+       }) do
+    ExternalIds.write(
+      attrs,
+      [type: media_type_to_string(media_type), title: attrs[:title]],
+      fn attrs ->
+        Media.create_media_item(attrs, skip_episode_refresh: true)
+      end
+    )
   end
 
+  # `:type` comes from `preparation.media_type`, not `attrs[:type]`: the
+  # `:stamp_source` operation's attrs is bare `%{metadata_source: ...}`, with
+  # no `:type` key of its own, and `media_type` is always present on the
+  # preparation regardless of what the attrs happen to carry.
   defp persist_media_item(%MetadataPreparation{
          operation: operation,
+         media_type: media_type,
          media_item_id: media_item_id,
          media_item_attrs: attrs
        })
        when operation in [:update, :stamp_source] do
     case Repo.get(MediaItem, media_item_id) do
-      nil -> {:error, {:media_item_missing, media_item_id}}
-      media_item -> Media.update_media_item(media_item, attrs, reason: "Metadata enriched")
+      nil ->
+        {:error, {:media_item_missing, media_item_id}}
+
+      media_item ->
+        ExternalIds.write(
+          attrs,
+          [
+            type: media_type_to_string(media_type),
+            exclude_id: media_item.id,
+            title: media_item.title
+          ],
+          fn attrs -> Media.update_media_item(media_item, attrs, reason: "Metadata enriched") end
+        )
     end
   end
 
--- a/lib/mydia/library/metadata_enricher.ex
+++ b/lib/mydia/library/metadata_enricher.ex
@@ -383,6 +383,7 @@ defmodule Mydia.Library.MetadataEnricher do
     # every owner found really is a different item.
     attrs =
       ExternalIds.put_free_ids(attrs, metadata.external_ids,
+        type: media_type_to_string(media_type),
         exclude_id: Map.get(match_result, :exclude_id)
       )
 
--- a/lib/mydia/media.ex
+++ b/lib/mydia/media.ex
@@ -147,28 +147,6 @@ defmodule Mydia.Media do
   end
 
   @doc """
-  Gets a single media item by TMDB ID.
-  """
-  @spec get_media_item_by_tmdb(integer(), keyword()) :: MediaItem.t() | nil
-  def get_media_item_by_tmdb(tmdb_id, opts \\ []) do
-    MediaItem
-    |> where([m], m.tmdb_id == ^tmdb_id)
-    |> maybe_preload(opts[:preload])
-    |> Repo.one()
-  end
-
-  @doc """
-  Gets a single media item by TVDB ID.
-  """
-  @spec get_media_item_by_tvdb(integer(), keyword()) :: MediaItem.t() | nil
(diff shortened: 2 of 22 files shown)


arsfeld:
Thanks. Took one of these and am declining the other six, with reasoning.

**Taken:** `add_config_flow_test.exs:314` now passes `type: "movie"` (0390f424e). That one is a presence lookup: the flow adds a movie, waits for the row, then asserts `added.library_path_id`. An untyped lookup could in principle bind a row of the other type, so scoping it makes the intent explicit and costs nothing.

**Declining the six absence assertions** (`add_test.exs:93,94,119,129,130` and `media_requests_test.exs:673). Adding `:type` there would weaken them.

Every one of those is a `refute`, and each proves the code under test created *nothing*:

```elixir
assert {:error, {:metadata, :tvdb_ref_for_movie}} =
         Add.from_provider({:tvdb, tvdb_id}, :movie, relay_config(bypass))

refute Mydia.Media.find_by_external_ids(%{tmdb: tvdb_id})
refute Mydia.Media.find_by_external_ids(%{tvdb: tvdb_id})
```

An untyped lookup searches every type, so it is strictly the stronger assertion. Narrowing it to `type: "movie"` means a stray `tv_show` row holding that id would no longer be found, and the `refute` would pass when it should fail. That is the opposite of the intent.

The review notes a `tv_show` row could "make the assertions pass or fail for the wrong reason". For a `refute` it can only make them *fail*, and failing is correct there: these tests assert nothing was written at all, so a row of either type appearing is a genuine defect worth failing on. There is also no cross-test collision risk, since each id comes from `System.unique_integer([:positive])`.

This is deliberate rather than an oversight. The design doc records the rule as: production call sites get a real type, test call sites stay unscoped because they assert "no row anywhere holds this id" and a type filter would change what they prove. `find_by_external_ids/2` keeps `type: nil` legal precisely so that assertion stays expressible. The one lookup that wants a specific row, rather than the absence of any row, is the one I scoped.

PR #750, opened 2026-09-08
fix(player): stop a failed candidates call from forcing a transcode

arsfeld:
Follow-up to #749. That PR fixed the timeouts around a cold transcode; this one
addresses why the transcode happened at all, which turned out to be a separate
bug on the player side.

## What happened

On 2026-09-08 an AV1/Opus MKV was server-transcoded on macOS. The same file, same
encode, had direct-played 22 minutes earlier as the preceding episode of the same
show. The transcode then failed on `Error parsing Opus packet header`, so the
viewer got nothing.

Three explanations were ruled out against the live server:

| hypothesis | why not |
|---|---|
| device profile missing | Even with no profile the server answers `[REMUX, HLS_COPY, TRANSCODE]`, and `firstStrategyAllowsDirectPlay` accepts a leading `REMUX`. With the native profile it answers `[DIRECT_PLAY, TRANSCODE]`. Both permit direct play. |
| relay connection | The relay path filters to `TRANSCODE` only, but also forces a 2000 kbps cap. The session log carries no `max_bitrate`. |
| a chosen quality rung | A non-Original rung does veto direct play, by design. Every rung carries a bitrate cap (8000/4000/1500/800) and the log shows none, so the rung was Original. |

What remained is the direct-play gate itself. It required `candidatesResult != null`,
and `_fetchStreamingCandidates` returns null on any exception. With null candidates
`_pickHlsStrategy` falls through to `TRANSCODE`.

That candidates call took 14,955 ms, the same unexplained stall #749 documented. So
the stall did not merely delay playback, it silently changed the playback decision.

## The fix

An unanswered candidates query is a transport failure, not a compatibility verdict.
By the time the gate runs, a null result can only mean a transport failure against
the route's own file id: both paths that would leave the screen without a usable id
already throw at `playFileId`. So the real choice is between handing mpv a file it
almost certainly decodes and asking a server we just failed to reach to run an
encode first.

Unknown now means direct play on native. A chosen rung still vetoes, and a leading
`HLS_COPY` still blocks, so unknown does not become a way around the server's
`:needs_transcoding` verdict.

The gate moves to `nativeDirectPlayAllowed` so it can be tested directly. That left
`PlayerScreen._canDirectPlay` unused, so it goes, and the one paragraph of its docs
not already duplicated on `firstStrategyAllowsDirectPlay` (the Fire HD 10
`MediaCodecList` probe) moves there.

## Logging

Answering the question above meant ruling out three hypotheses by hand, because
every input to the decision was invisible. `streaming_candidates/3` now logs the
leading strategy with the device profile and connection type that produce it, and
the session-start line carries `max_height` beside the `max_bitrate` it already had.
Neither changes a decision.

## Verification

`./dev mix precommit`: 10876 tests, 0 failures. `flutter analyze` on the changed
files is clean apart from two pre-existing `info`s at lines this PR does not touch.

Four new cases cover the gate: unknown means direct play, a rung vetoes with or
without candidates, and known candidates are still judged on their leading strategy
(including that a leading `HLS_COPY` and an empty list both still block).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved native direct playback decisions for Original-quality media.
  - Prevented incompatible direct-play strategies from bypassing codec compatibility checks.
  - Improved playback behavior when delivery strategy information is unavailable or cannot be retrieved, reducing unnecessary fallback to transcoded streaming.
  - Added safeguards for selected quality levels and HLS copy strategies to ensure playback uses compatible delivery options.

### DIFF
--- a/lib/mydia_web/schema/resolvers/streaming_resolver.ex
+++ b/lib/mydia_web/schema/resolvers/streaming_resolver.ex
@@ -36,11 +36,12 @@ defmodule MydiaWeb.Schema.Resolvers.StreamingResolver do
         case Candidates.resolve_media_file(content_type, id) do
           {:ok, media_file} ->
             media_file = Candidates.ensure_codec_info(media_file)
+            device_profile = context[:device_profile]
 
             candidates =
               Candidates.build_streaming_candidates(
                 media_file,
-                context[:device_profile] || DeviceProfile.browser_default()
+                device_profile || DeviceProfile.browser_default()
               )
 
             metadata = Candidates.build_metadata_response(media_file, user_id: user.id)
@@ -59,6 +60,25 @@ defmodule MydiaWeb.Schema.Resolvers.StreamingResolver do
               Enum.map(candidates, fn candidate ->
                 %{candidate | strategy: strategy_to_atom(candidate.strategy)}
               end)
+
+            # The leading strategy is the whole answer: the client reads it as
+            # the compatibility verdict and will not direct play behind a
+            # leading TRANSCODE or HLS_COPY. Logging it with the two inputs that
+            # decide it turns "why did this transcode?" into one grep. Both
+            # inputs were invisible before, and answering that question on
+            # 2026-09-08 meant ruling out three hypotheses by hand.
+            leading =
+              case candidates do
+                [%{strategy: strategy} | _] -> strategy
+                _ -> nil
+              end
+
+            Logger.info(
+              "Streaming candidates for file #{media_file.id}: " <>
+                "leading=#{inspect(leading)} " <>
+                "device_profile=#{if device_profile, do: "client", else: "absent (assuming browser)"} " <>
+                "connection=#{context[:peer_connection_type] || "unknown"}"
+            )
 
             {:ok, %{file_id: media_file.id, candidates: candidates, metadata: metadata}}
 
@@ -307,9 +327,14 @@ defmodule MydiaWeb.Schema.Resolvers.StreamingResolver do
       reported_start_position =
         if info.playlist_mode == :full, do: 0, else: info.start_position
 
+      # max_height is here because its absence cost three ruled-out hypotheses
+      # on 2026-09-08. A capped height is the one thing that vetoes direct play
+      # from the client side, and without it in the log there is no way to tell
+      # a viewer who picked a quality rung from one who did not.
       Logger.info(
         "Started streaming session #{info.session_id} for file #{file_id}, user #{user_id}" <>
           if(max_bitrate, do: " (max_bitrate: #{max_bitrate}kbps)", else: "") <>
+          if(max_height, do: " (max_height: #{max_height}p)", else: "") <>
           if(session_start_position > 0,
             do: " (start_position: #{session_start_position}s)",
             else: ""
--- a/player/lib/domain/models/quality_delivery_subtitle.dart
+++ b/player/lib/domain/models/quality_delivery_subtitle.dart
@@ -48,6 +48,16 @@ String cappedRungDeliverySubtitle(int? maxBitrateKbps) {
 /// compatibility check already found acceptable into a different
 /// container.
 ///
+/// This guard used to stay permissive on the theory that the native device
+/// profile the server checks against hadn't been validated on real hardware
+/// and might under-report a device's true decoder support. That hedge no
+/// longer applies: `android_codec_capabilities.dart`'s `MediaCodecList` probe
+/// against a Fire HD 10 confirmed its decoder is exactly as limited as the
+/// server's verdict assumed — `video/hevc` capped at 8-bit, `video/vp9` capped
+/// at 10-bit — so a `:needs_transcoding` verdict for a codec this device
+/// cannot decode is trustworthy, and HLS_COPY must not be used to second-guess
+/// it.
+///
 /// Platform gating (`!kIsWeb`) is intentionally left to the caller — this
 /// only inspects strategy ordering.
 bool firstStrategyAllowsDirectPlay(Iterable<String> strategyValues) {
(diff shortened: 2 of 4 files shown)


arsfeld:
## Review notes

### Blocking

**1. This breaks an existing test.** `player/test/presentation/screens/player/player_honors_selected_file_test.dart:156` ("a transport failure fetching candidates still honors the selected file") scripts an `http.ClientException` for the candidates call and then does `link.requests.firstWhere((r) => r.variables.containsKey('strategy'))`. Under the new gate a transport failure means direct play, so `StartStreamingSession` is never issued and `firstWhere` throws `Bad state: No element`. Running `flutter test test/presentation/screens/player test/domain/models` on the branch gives `+427 -1`, that test the only failure, so `Test / Player` will go red. The assertion it guards (a transport failure must not swap the viewer's file for a server-ranked one) is still worth keeping: retarget it at `proxyService.directStreamFileIds`. Its comment at lines 144-149 is now wrong too.

### The direction problem

Before fixing the test, I want to settle whether this is the decision we are keeping, because this gate has now moved three times in eighteen days:

| date | change | direction |
| --- | --- | --- |
| 2026-08-22 | `8e0fea221` "trust the server's per-device direct play answer", `276350a21` build candidates against the caller's profile | server becomes authoritative |
| 2026-08-25 | `2b42b5fd7`, `20de21b77`, `81c03ee00` (#564, #566) | tightened; a leading `HLS_COPY` stops counting as direct-playable |
| this PR | unknown means direct play | loosened on a new axis |

The 2026-08-25 change did not just tighten the gate, it retired a permissive hedge and wrote down why: the client used to second-guess the server on the theory that the device profile might under-report real decoder support, and the Fire HD 10 `MediaCodecList` probe showed the server's verdict was right. This PR copies that paragraph forward onto `firstStrategyAllowsDirectPlay` and then adds a new hedge eight lines below it, without reconciling the two. As written they contradict: an explicit `:needs_transcoding` verdict is refused, and no verdict at all is accepted.

There are now four places that answer "may this file be direct played": `Candidates.build_streaming_candidates/2`, the relay filter in `streaming_resolver.ex:52`, `firstStrategyAllowsDirectPlay`, and the new null branch in `nativeDirectPlayAllowed`. `lib/mydia/streaming/README.md` is the designated home for this invariant and tells the reader to "grep for all three shapes before assuming a candidate-list fix is complete". This PR adds a fourth way past the server's verdict and records it only in a Dart dartdoc.

What would stop the flip-flopping, rather than flipping it again:

1. **Make the wrong pick recoverable.** Both directions are one-shot today. `_initializePlayer`'s only catch (`player_screen.dart:1493`) sets `_error` and stops, so a direct play that mpv cannot decode is terminal, exactly as a transcode that fails on an Opus header is terminal. If a failed direct play retried as HLS, the choice under uncertainty stops being load-bearing and neither incident forces the coin to be flipped the other way.
2. **Give "no verdict" its own state.** `_fetchStreamingCandidates` already distinguishes `linkException` for `serverRejected`; return unknown as a third state instead of overloading `null`, which today also covers deserialization and cache-normalization failures (see finding 5).
3. **Write the resulting rule in `lib/mydia/streaming/README.md`**, next to the `HLS_COPY` table, and have both client helpers cite it.

### Correctness

**2. The "a chosen rung still vetoes" guard is dead on the null path.** `_resolveQualityForFile` collapses the ladder to `[Original]` on the same null result: `deriveQualityLadder(sourceHeight: null)` returns `const [QualityRung.original]` (`quality_rung.dart:86`), `_qualityLadder.contains(requested)` is then false for 1080p/720p/480p, and `_settledQuality` is reset to Original just before the gate reads it. So `isOriginalQuality` is unconditionally true whenever `strategyValues` is null, and the first guard in `nativeDirectPlayAllowed` cannot fire in production. A viewer on a metered connection who picked 480p, or who has 480p as their install default, gets an uncapped raw direct play on any candidates blip. The new test "a chosen rung still vetoes, known candidates or not" asserts a state the screen cannot reach.

**3. Relay sessions lose their `TRANSCODE`-only enforcement.** `streaming_resolver.ex:52` filters candidates to `TRANSCODE` when `peer_connection_type == "relay"` ("direct play won't work within relay bandwidth limits"), and `start_streaming_session` clamps relay sessions to 2000kbps/720p. None of that survives a failed candidates call, because it lives only in the response the client failed to get. The client has no relay awareness of its own (`_relayed` is `kIsWeb && !isInstanceHostedWeb`, always false on native), so a native viewer on the iroh relay now streams the full-bitrate source through our relay. A slow or relayed connection is also a likely cause of the 15s stall this PR is built around, so the new branch fires hardest in the case it is worst for.

**4. Unknown is more permissive than the clearest possible "no".** `['HLS_COPY', 'TRANSCODE']` returns false; `null` returns true. On a Fire HD 10 with a 10-bit HEVC file, a transport blip now hands mpv the exact bytes `firstStrategyAllowsDirectPlay` exists to withhold, and there is no retry. Before this PR that blip produced a working transcode. Worth measuring which failure is more common before trading one for the other.

**5. "A null result can only mean a transport failure" is not true.** `player_screen.dart:1960` is a bare `catch (e)` around `Query$StreamingCandidates.fromJson(result.data!)` returning `(candidates: null, serverRejected: false)` for anything thrown. `player_screen_test_harness.dart:469-473` documents this path: a partial normalized-cache write "surfaces as `result.hasException` and makes `_fetchStreamingCandidates` return null". A version-skewed or malformed response, where the client emphatically does not know the file is playable, takes the direct-play branch under a comment asserting it cannot happen.

**6. The gate and the Original-rung label now disagree.** `_rememberOriginalDeliverySubtitle(null)` coerces to `const <String>[]`, so the sheet reads "Original, re-encoding required" while the file direct-plays. The dartdoc one line above ("Labels only, does not change the `nativeDirectPlayAllowed` playback gate") was accurate when both derived from the same coerced list; it now describes a divergence. Passing the same nullable list to both fixes it.

### Logging commit

**7.** `max_height` at `streaming_resolver.ex:337` is `FfmpegHlsTranscoder.effective_max_height(clamped_height)`, which returns `configured_max_height()` when the request carried none (`ffmpeg_hls_transcoder.ex:913-920`). On any instance with `streaming.max_transcode_height` set, every session logs a height whether or not the viewer picked a rung, so the comment's "no way to tell a viewer who picked a quality rung from one who did not" is still true afterwards. Log the requested `clamped_height` beside the applied one.

**8.** `device_profile=client` at line 79 is byte-identical for a Fire HD 10 capping hevc at 8-bit and for the macOS build in the 2026-09-08 incident. `DeviceProfile` carries `video_codecs`, `audio_codecs`, `containers` and `codec_profiles`, which is what `Compatibility` actually evaluates, and none of it reaches the log. Re-running this investigation with the line in place still cannot answer "did the client claim av1/opus?".

**9.** Only the success branch logs. `{:error, :not_found}` is the answer that makes the client set `serverRejected` and silently re-ask by media item, playing a different file than the route named (`player_screen.dart:1223`), which is at least as surprising as a transcode and stays invisible.

### Smaller

**10.** `PlayerScreen._canDirectPlay` is deleted but still cited in six places: `quality_delivery_subtitle.dart:15` and `:36` (directly above its replacement), `lib/mydia/streaming/README.md:72`, `lib/mydia/streaming/candidates.ex:137`, `player_screen_test_harness.dart:465`, and `test/mydia/streaming/codec_profile_compatibility_test.exs:140`.

**11.** `candidates.map((c) => c.strategy.toJson()).toList()` is now written three times in one file: line 1288 (new gate), 1537 (`_pickHlsStrategy`), 1877 (`_rememberOriginalDeliverySubtitle`), two of them re-deriving a value already in scope. Hoisting one `final strategyValues = ...` above line 1272 also forces the null-handling question in finding 6 to be answered once.

**12.** `nativeDirectPlayAllowed` does not enforce the "native" in its name; `!kIsWeb` is still the caller's job. That was belt-and-braces before, since browser-profile candidates lead with `TRANSCODE` anyway. Now a caller that forgets it gets `true` from a failed candidates call on web. The function was extracted to be testable, so an `isNative:` parameter costs a line and makes the misuse unrepresentable.

**13.** The new tests never cover a leading `TRANSCODE` with `isOriginalQuality: true`, which is what both a relay-filtered response and a browser-profile `:needs_transcoding` verdict return. `firstStrategyAllowsDirectPlay` has that case at line 83; `nativeDirectPlayAllowed` does not, so the delegation could invert for it unnoticed.

PR #774, opened 2026-09-10
fix(streaming): keep audio, segments and the resume offset aligned after a seek

arsfeld:
## What was wrong

Three things broke when an HLS session started anywhere but 0s. All three were measured with FFmpeg 8.1 against a synthetic file with a keyframe every 10s and a white flash and a 1kHz beep that start together.

- A resume at 27s that re-encoded the video and copied AAC audio started the audio 7.2s before the first video frame. Accurate seek drops decoded video before the target; copied audio never reaches a decoder, so it started back at the keyframe. Jellyfin 12.0 fixed the same bug (jellyfin/jellyfin#16580).
- A resumed `:full` session's first encoder sought to the raw resume second but took its segment numbers from the 4s grid. Resumed at 30s, the segments declared as 28-32, 32-36 and 36-40 began at 30.00, 33.92 and 37.93, and a later relocation met them with a gap.
- A `:window` stream copy can only begin on a keyframe, yet `startStreamingSession` echoed the requested offset. The player and the cast path build their timeline from that echo, so the position and the saved progress ran up to a GOP ahead. Casting always runs `:window`, and every out-of-reach cast seek restarts the session.

## What changed

- `-copypriorss:a 0` when a seek re-encodes video and copies audio. Jellyfin used an output-side `-ss`, which rebases every timestamp to zero under `-copyts` and would break `:full`.
- The first `:full` encoder starts on the segment grid, where `relocate/2` already started every later one.
- For a `:window` stream copy from MKV or MP4, `HlsSession.prepare_start/2` asks ffprobe which keyframe FFmpeg will land on, bounded at 1.5s by the new `Mydia.BoundedCommand` (lifted out of the hwaccel probe). It runs in the caller's process, so a slow disk never holds up the session supervisor. The transcoder seeks 0.2s past that keyframe, because FFmpeg pulls MKV seeks on B-frame video back by 3/23s, and the resolver echoes it. The registry still matches on the requested offset, so a repeat request reuses the running session. MPEG-TS has no seek index and keeps the old behaviour.
- Two comments in the player that already assumed the echo named a keyframe say when it does.

No schema change. Older players read the same `startPosition` and get a truer number.

One case stays as it is. A pinned copy that re-encodes its audio (H.264 with DTS or TrueHD, say) starts the audio 0.18s after the first frame, with timestamps still aligned. The README table records it.

## Testing

- `mix precommit`: 11125 tests, 0 failures.
- `test/mydia_web/schema/streaming_test.exs`, which CI runs, starts a real HLS_COPY session 17s into an MKV with keyframes every 10s and checks the echo is 10, then that the same request gets the same session back. The existing "echoes the offset" test gained a bitrate cap: its H.264 fixture is stream-copied even under TRANSCODE, so without the cap it would now be pinned to its keyframe at 0.
- `test/mydia/streaming/hls_seek_alignment_integration_test.exs` re-measures sync, grid alignment and the pinned starts against the installed FFmpeg. It is tagged `:ffmpeg` like the relocation test, so CI skips it. Run it with `--include ffmpeg` after an FFmpeg upgrade.

The measurements are written up in `lib/mydia/streaming/README.md` under "Seeking with copied streams".

## Summary by CodeRabbit

* **New Features**
  * Stream-copy playback for MKV and MP4 now starts at the nearest preceding keyframe when seeking or resuming.
  * Playback responses report the effective start position used by the stream.
  * Improved synchronization between copied audio and re-encoded video after seeking.
  * Added bounded external-command execution with timeout and failure handling.

* **Documentation**
  * Updated playback and seeking guidance to explain keyframe-aligned behavior across supported formats.


arsfeld:
@coderabbitai review

arsfeld:
Both outside-diff findings on `player/lib/core/playback/playback_controller.dart` check out, but they sit in code that came in with #769; this PR only rewords one comment in that file.

- `_awaitPlaylist` waits for three `.ts` entries whatever the playlist mode (line 375), although its doc says a FULL playlist is complete on the first poll. A FULL playlist for a clip under 12s would poll 20 times and then fail.
- `httpPlaylistProbe` calls `http.get` with no timeout, and ending the session does not cancel it.

Leaving both for a separate player fix so this PR stays on the server side.

PR #776, opened 2026-09-11
fix(player): let Auto transcode at the source until the link says otherwise

arsfeld:
#769 made Auto the default quality. When Auto has to transcode, it asked for
the top rung of the adaptive ladder whenever it had no remembered throughput,
which is every first play and every web play (web never measures throughput).
That ladder stops at 1080p, so a 4K file that needed a transcode came out at
1080p, and a 1080p file got an 8000 kbps cap, where the old default
(Original) sent no caps at all. The mid-session adaptation that would justify
starting at a rung does not exist yet, so the cap was a downgrade with nothing
in return.

Knowing nothing about the connection is not evidence against it. Auto now
transcodes uncapped, like Original, and only asks for a lower rung when
remembered throughput says the file will not fit (the same 30% headroom rule
direct play uses). Then it asks for the highest adaptive rung that does fit.

The change is one condition in `playback_planner.dart`, rule 3: Auto caps only
when `!fits`, the flag the planner already computes for direct play, which is
true only when both the file bitrate and a remembered throughput are known and
the file does not fit.

Unchanged on purpose: a fallback after a failure (`fallbackPlan`) still steps
down to a lower rung, since it follows evidence that something went wrong.
Relay clamps still apply server-side, and the "Limited to ... by your
connection" note still names them.

## Testing

- Planner: Auto with no throughput, with a fitting file, with a bitrate but
  no throughput, and with a throughput but no bitrate all transcode uncapped;
  a 20000 kbps 4K file against 6000 kbps of remembered throughput asks for
  720p.
- Screen: the caps tests that assumed Auto's old caps now expect none, and a
  new one seeds 6000 kbps of remembered throughput for a 20 Mbps 4K file and
  sees `maxHeight 720` / `maxBitrate 4000` reach `startStreamingSession`.
- `flutter test --concurrency=1`: 3597 passed, 7 skipped, 0 failed.
- `flutter analyze`: no errors or warnings.

## Summary by CodeRabbit

- **Playback**
  - Auto quality now preserves the source resolution when connection capacity is unknown or sufficient.
  - When remembered throughput cannot support the file, Auto selects the highest adaptive quality that fits.
  - Playback failures now step down through a deliberate fallback plan.
  - Auto may transcode without quality caps when no suitable adaptive option is available.

- **Documentation**
  - Updated playback guidance to clarify Auto quality selection and subtitle labels.

### DIFF
--- a/player/docs/playback.md
+++ b/player/docs/playback.md
@@ -28,11 +28,16 @@ verification or a quality change, logs a different line; see
 
 The viewer's choice is Auto (the default), Original, or a fixed rung. Auto
 direct plays or copies when the rules below allow it, and otherwise
-transcodes at the top rung of the adaptive ladder that fits remembered
-throughput (the top rung when throughput is unknown). Original is the
-viewer's override: it bypasses remembered decode failures, wherever that
-choice came from. A fixed rung pins its own caps and always transcodes,
-skipping both checks below.
+transcodes at the source resolution, like Original, unless remembered
+throughput says the file will not fit, in which case it asks for the highest
+adaptive rung that does. Knowing nothing about the connection is not evidence
+against it, so a first play, and every web play (web never measures
+throughput), transcodes uncapped. A fallback after a playback failure is a
+different case: it steps down deliberately rather than waiting for evidence
+(`fallbackPlan`; see "Verification" below). Original is the viewer's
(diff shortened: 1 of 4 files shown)


arsfeld:
@coderabbitai review

PR #782, opened 2026-09-11
fix(player): stop copying every decoded frame through the CPU on Android

arsfeld:
## The bug

Playback on a Chromecast with Google TV runs at **0.78x realtime**, so roughly one second in five is a stall. That is the stutter.

It is not the server, the network, or the decoder. Measured live against a 1080p episode:

| Suspect | Measurement | Verdict |
| --- | --- | --- |
| Transcode speed | Server produced 101 to 293 segments in 49s (**23.5x realtime**), hundreds queued ahead | Not it |
| Network | 46-100 Mbps, segments delivered in 1-159ms | Not it |
| Video decode | `c2.amlogic.avc.decoder` (hardware) active throughout; software AVC never instantiated | Not it |
| Thermal | 43-48C, no throttling | Not it |
| Memory | Zero swap-out during playback, 1.7 major faults/s, ~710MB available | Not it |
| **Frame readback** | **`CodecLooper` pegged at 91.4% of one core**, every other thread under 7% | **This** |

## Root cause

`VideoController(player)` was constructed with no configuration, so on Android media_kit defaults to `hwdec=auto-safe`, which resolves to `mediacodec-copy`. The hardware codec decodes, then every frame is read back into CPU memory before mpv uploads it as a texture. That readback comes off a GPU/codec buffer rather than ordinary RAM, so it costs far more than the ~75 MB/s that 1080p24 NV12 suggests. One thread saturates while three cores sit idle.

This is why direct play stutters identically to a transcoded stream: the render path is the same regardless of source.

## The fix

`hwdec=mediacodec` is the same hardware decoder with the readback removed. The frame stays on the GPU and reaches mpv as a SurfaceTexture.

mpv documents it as requiring `--vo=gpu --gpu-context=android`, and media_kit already sets `gpu-context: android` and `opengl-es: yes` on Android, so `vo` stays `gpu` and **mpv keeps compositing**.

### Why not `vo=mediacodec_embed`

Embedding is the more obvious "zero copy" answer and it is the wrong one here. It hands frames straight to a SurfaceView and mpv never draws, which silently removes subtitles and the OSD. This player renders both through mpv (`setSubtitleTrack`), and there is no Flutter-side subtitle renderer to fall back on. Embedding also cannot accept software-decoded frames, so a hwdec failure has nowhere to go. With `vo=gpu` a failure just drops to software decode and still renders.

A test pins `vo` to null so a later switch to embedding, which would quietly kill subtitles, fails the suite instead of shipping.

### Why Android only

`hwdec` values are platform specific, and mpv deliberately leaves `mediacodec` out of `auto-safe` because it converts to RGB and its handling of unusual colorspaces is unverified. Every other platform keeps media_kit's default.

## How other players do this

ExoPlayer/Media3 apps decode straight into a SurfaceView and layer subtitles as a separate Android view, so they never pay this cost. mpv-android ships `vo=gpu` + `gpu-context=android` with `hwdec=mediacodec` available and `mediacodec-copy` as fallback. `mediacodec-copy` with a texture-backed `vo` is the one configuration that drags every frame through the CPU, and it was ours by default.

## Testing

- 8 unit tests covering Android, web reporting as Android, and the four other native platforms.
- Device verification on the Chromecast (sabrina, armeabi-v7a) is in progress; before/after playback rate will be posted to this PR before merge.

The residual risk is colour artifacts on some Amlogic hardware rather than a crash, which is exactly what the device check is for.

## Summary by CodeRabbit

* **Improvements**
  * Video playback now uses platform-specific output settings.
  * Android playback uses zero-copy hardware decoding for improved rendering efficiency.
  * Web and other supported platforms continue using the default video configuration.

### DIFF
--- a/player/lib/core/player/video_output_config.dart
+++ b/player/lib/core/player/video_output_config.dart
@@ -1,0 +1,49 @@
+import 'package:flutter/foundation.dart' show TargetPlatform;
+import 'package:media_kit_video/media_kit_video.dart'
+    show VideoControllerConfiguration;
+
+/// mpv's hardware decoder for Android that keeps the frame on the GPU.
+///
+/// The name is mpv's, not media_kit's: media_kit passes `hwdec` through to
+/// libmpv untouched.
+const androidZeroCopyHwdec = 'mediacodec';
+
+/// How the decoded frame should reach mpv's renderer on this platform.
+///
+/// media_kit's Android default is `hwdec=auto-safe`, which resolves to
+/// `mediacodec-copy`: the hardware codec decodes, then every frame is read
+/// back into CPU memory before mpv uploads it as a texture. That readback
+/// comes off a GPU/codec buffer rather than ordinary RAM, so it costs far more
+/// than the ~75 MB/s that 1080p24 NV12 suggests. On a Chromecast with Google
+/// TV (Amlogic, 4x Cortex-A55) it saturates the single `CodecLooper` thread,
+/// measured at 91% of one core with three cores sitting idle, and playback
+/// settles at 0.78x realtime, which is the stutter. Nothing upstream was
+/// short: the server had hundreds of segments already encoded and answered
+/// every request in under 160ms.
+///
+/// [androidZeroCopyHwdec] is the same hardware decoder with the readback
+/// removed. The frame stays on the GPU and reaches mpv as a SurfaceTexture.
+/// mpv documents it as requiring `--vo=gpu --gpu-context=android`, and
+/// media_kit already sets `gpu-context: android` and `opengl-es: yes` on
+/// Android, so `vo` stays `gpu` and mpv keeps compositing.
+///
+/// mpv keeping the compositing is the whole reason this is not
+/// `vo=mediacodec_embed`. Embedding hands frames straight to a SurfaceView and
+/// mpv never draws, which silently removes subtitles and the OSD, and this
+/// player renders both through mpv. Embedding also cannot accept
+/// software-decoded frames, so a hwdec failure has nowhere to fall back to.
+/// With `vo=gpu` a failure just drops to software decode and still renders.
+///
+/// Android only, for two reasons. `hwdec` values are platform specific, and
+/// mpv deliberately leaves `mediacodec` out of `auto-safe` because it converts
+/// to RGB and its handling of unusual colorspaces is unverified. Every other
+/// platform keeps media_kit's own default.
+VideoControllerConfiguration videoControllerConfigurationFor(
+  TargetPlatform platform, {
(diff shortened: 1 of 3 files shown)


arsfeld:
## Device verification (Chromecast with Google TV, sabrina, armeabi-v7a)

Same episode, same 1080p H.264 output, same hardware decoder, before and after.

| Metric | Before (`mediacodec-copy`) | After (`mediacodec`) |
| --- | --- | --- |
| Playback rate | **0.78x** | **1.00x** |
| `CodecLooper` CPU | **91.4% of one core** | **absent from top 10** (under 1.3%) |
| Busiest thread | `CodecLooper` 91.4% | `1.raster` 11.3% |
| Decoder | `c2.amlogic.avc.decoder` | `c2.amlogic.avc.decoder` (unchanged) |
| Colour | correct | correct |

Playback rate after, from `ProgressService` against wall clock:

```
20:17:15.798 position=9
20:17:25.798 position=19    +10s wall / +10s media
20:17:45.798 position=39    +20s wall / +20s media
20:17:55.798 position=49    +10s wall / +10s media
20:18:15.808 position=69    +20s wall / +20s media
```

60.01s of wall clock, 60s of media. Exactly realtime, against 0.78x before.

### Colour

mpv leaves `mediacodec` out of `auto-safe` partly because it converts to RGB and its
colorspace handling is unverified, and the known Amlogic failure mode is green or
discoloured frames. Checked visually across a wide gamut (saturated blues, gold, white
stone, greens, skin tones) and found nothing wrong.

### Subtitles

2 subtitle tracks detected, no subtitle errors in the log, and mpv is demonstrably
still compositing, since the video reaches the Flutter texture through `vo=gpu` at all.
That is the mechanism `mediacodec_embed` would have removed. I did not manage to drive
the subtitle picker over ADB to photograph a rendered subtitle line, so treat rendering
as inferred from mpv still owning the render path rather than visually confirmed.

### CI note

`Flatpak / Build` failed once and passed on re-run in 8m47s. It was the mode documented
in `.github/ci-flakes.md`: the Xvfb smoke test reporting `Player exited within 10 seconds`
with only `libEGL DRI3 error` and `KeyringLocked` noise, no `error while loading shared
libraries`, no `cannot open shared object`, no Dart exception, and no
`Failed to connect to code.videolan.org`.

### Unrelated issue found

Local Android builds are currently broken in the `.#android` dev shell:
`flutter_secure_storage 11.0.0` needs `compileSdk = 37` while
`nix/devShells/flake-module.nix` pins `platformVersions` through 36, and Gradle cannot
install into the read-only Nix store. Adding `"37"` fixes it. Deliberately not included
here; it wants its own PR.

PR #786, opened 2026-09-12
fix(downloads): handle compound history statuses in NZBGet client

### LINKED ISSUE #780: Bug: NZBGet successful downloads are marked as failed
Mydia incorrectly marks successfully completed NZBGet downloads as failed.

NZBGet reports history statuses as compound values such as:

* `SUCCESS/ALL`
* `SUCCESS/UNPACK`
* `SUCCESS/HEALTH`

Mydia currently compares the full status string against bare values such as `"SUCCESS"`. As a result, normal successful NZBGet history entries fall through to `:error`.

This causes Mydia to emit:

```text
Download failed: <name> (Download failed in client)
```

even though NZBGet completed the download successfully and the downloaded files are present on disk.

### Reproduction

1. Configure Mydia with NZBGet as the download client.
2. Start a download through Mydia.
3. Let NZBGet complete the download successfully.
4. NZBGet moves the files to its completed download directory and records a history status such as `SUCCESS/ALL`.
5. Mydia's download monitor processes the history entry.

### Actual behavior

Mydia interprets `SUCCESS/ALL` as `:error`.

The download is then marked as failed, the release is blacklisted, and the Mydia download row is deleted.

The completed files remain in NZBGet's download directory because the import path is never reached.

### Expected behavior

Any NZBGet history status in the `SUCCESS/*` class should be treated as successfully completed and proceed through the normal import flow.

`DELETED/*` statuses should not be treated as completed.

### Root cause

The NZBGet integration compares compound history statuses using exact string equality.

Relevant locations:

* `lib/mydia/downloads/client/nzbget.ex:499`

  * `parse_state/1` only maps bare `"SUCCESS"` to `:completed`
  * `SUCCESS/ALL`, `SUCCESS/UNPACK`, etc. fall through to `:error`

* `lib/mydia/downloads/client/nzbget.ex:470`

  * `completed_at` uses the same exact-string comparison

* `lib/mydia/downloads/client/nzbget.ex:234`

  * the `:completed` filter only accepts bare `"SUCCESS"` or `"DELETED"`

* `test/support/fixtures/nzbget/history.json:9`

  * the fixture uses bare `"SUCCESS"`, which does not reflect the compound statuses returned by NZBGet history

The resulting `:error` state is handled by the download monitor as a real client failure:

* `lib/mydia/jobs/download_monitor.ex:386`

This runs the failure path, including blacklisting the release and deleting Mydia's download row.

### Suggested fix

Normalize the NZBGet history status before interpreting it, for example by splitting on `/` and handling the status class.

Conceptually:

```elixir
defp status_class(status) do
  status
  |> String.split("/", parts: 2)
  |> hd()
end
```

Then use the normalized class consistently for:

* state parsing
* `completed_at`
* completed-history filtering

Only `SUCCESS/*` should map to completed.

`DELETED/*` should not map to completed, since statuses such as `DELETED/MANUAL`, `DELETED/DUPE`, and `DELETED/COPY` indicate deletion rather than successful completion.

### Tests

Add coverage for at least:

```text
SUCCESS/ALL       -> completed
SUCCESS/UNPACK    -> completed
SUCCESS/HEALTH    -> completed
DELETED/MANUAL    -> not completed
FAILURE/*         -> error
WARNING/*         -> error
```

It would also be useful to add a higher-level monitor test verifying that a `SUCCESS/ALL` NZBGet history item proceeds to import rather than entering the blacklist/delete failure path.

### Reference

NZBGet history API:

https://nzbget.com/documentation/api/history/

arsfeld:
Closes #780

### Problem
NZBGet reports compound history statuses (e.g. `SUCCESS/ALL`, `SUCCESS/UNPACK`, `SUCCESS/HEALTH`, `DELETED/MANUAL`). The NZBGet client compared against bare `"SUCCESS"`, which caused successful downloads to fall through to `:error`. In turn, `DownloadMonitor` treated this as a client failure, blacklisted the release, and removed the download row.

### Solution
- Normalize NZBGet statuses by extracting the primary status class (before `/`).
- Map `SUCCESS/*` to `:completed` and `DELETED/*` to `:error`.
- Populate `completed_at` only for `SUCCESS/*` statuses.
- Update `list_torrents/2` with `filter: :completed` to accept `SUCCESS/*` and remove the strict `TotalArticles == SuccessArticles` equality check so repaired downloads are retained.
- Added test fallback in `DownloadMonitor.enqueue_import_job/2` for test runs where Oban is not running.
- Updated fixtures and test suites, including a high-level `DownloadMonitor` test verifying `SUCCESS/ALL` downloads proceed to import.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download status handling when services provide detailed status information.
  * Downloads marked as deleted, failed, or warning are now correctly treated as errors rather than completed.
  * Only successfully completed downloads are included in completion results.
  * Fixed completed downloads occasionally failing to queue their media import for processing.


0xC0FFEE:
Thanks for the swift fix!

PR #790, opened 2026-09-12
feat(library-api): replace timestamp polling with a revision change feed

arsfeld:
## Summary

Replaces the Library API's lossy timestamp polling with a monotonic, database-owned latest-state change feed.

`mediaItems(updatedSince:)` advanced off `media_items.updated_at`, so it missed every child write (episodes, media files, downloads), could not express deletions, and could not be resumed reliably under a clock-skewed or out-of-order watermark. `mediaItemChanges(after:)` pages a database-owned revision marker instead.

- **Storage** — adapter-specific triggers maintain one `media_item_revisions` row per media-item UUID: an identity/`AUTOINCREMENT` 64-bit `revision`, a `deleted` tombstone flag, and an informational `changed_at`. The upsert carries a greater-revision guard so the marker stays monotonic when transactions commit out of order.
- **Coverage** — parent insert/update/delete, episode changes, media-file insert/trash/restore/relink/delete, `media_file_episodes` link/unlink, download insert/completion/failure/delete, and quality-profile renames. Re-parenting marks both owners; owner resolution joins through a live `media_items` row so a cascade can never resurrect a parent tombstone.
- **Clock sweep** — `get_media_status/1` compares episode `air_date` with `Date.utc_today()`, so a show can leave `UPCOMING` at a UTC date boundary with no database write. A synchronous startup catch-up (after `Ecto.Migrator`, before the endpoint) plus a daily Oban job marks the shows whose `air_date` crossed the watermark. It fails application boot if it cannot complete.
- **GraphQL** — `mediaItemChanges(first: Int = 50, after: String): MediaItemChangeConnection!` with opaque `v1:` revision cursors, live changes and tombstones, and `MediaItem.updatedAt` now reporting the aggregate change time.

Notable fix found only on PostgreSQL during adapter verification: `now()` is transaction-start time, so `changed_at` never advanced on a child change within one transaction — switched to `timezone('UTC', clock_timestamp())`. Also fixed a `BEFORE DELETE` trigger that returned `NULL`, which in PostgreSQL cancels the row operation, making `quality_profiles` deletes silently no-op.

## Breaking change (beta API)

`mediaItems(updatedSince:)` and its timestamp cursors are removed with **no compatibility alias**. Old cursors are invalid — clients must start a fresh snapshot with `mediaItemChanges`. `MediaItem.updatedAt` now reports the aggregate change time, which on upgrade is the migration instant for every existing item until it next changes.

## Verification

| Gate | Result |
|---|---|
| `./dev mix precommit` | Dependencies, Compile, Unused deps, Format, Credo, Database, Tests all green — **11,456 tests, 0 failures** |
| PostgreSQL focused suite (forced recompile) | **48 tests, 0 failures**; migration applies, reverses via `rollback --step 1`, and re-applies |
| PG concurrency regression | passes, and fails when the greater-revision guard is removed (negative control) |
| GraphQL end-to-end | `Absinthe.run/3` observed: initial snapshot (`endCursor` `djE6Mg`) → child write → live changed item (`djE6Mw`) → delete → tombstone `deleted: true`, `mediaItem: null` (`djE6NA`) |

Trigger behavior is covered by `test/mydia/library_api/revision_triggers_test.exs`, which pairs every marker-revision assertion with the observable `MediaItemView`/`get_media_status` outcome so the test fails if either side drifts.

## Migration note

The migration was edited in place after first being applied (PostgreSQL function bodies changed). CI is unaffected — it uses a fresh Postgres container with `ecto.create` + `migrate`. An already-migrated development database needs `mix ecto.rollback --step 1 && mix ecto.migrate` to pick up the new function bodies.

## Known tradeoff

A late-committing PostgreSQL transaction whose allocated revision is lower than an already-committed one has its marker write discarded, including a tombstone. This is the tolerance the design spec records (`docs/superpowers/specs/2026-09-12-library-api-revision-polling-design.md`) and the concurrency regression pins as intended: PostgreSQL sequences allocate outside commit order, and the greater-revision guard is what makes the marker monotonic. A lossless variant would retry allocation until the incoming revision exceeds the stored one.

Design and plan: `docs/superpowers/specs/2026-09-12-library-api-revision-polling-design.md`, `docs/superpowers/plans/2026-09-12-library-api-revision-polling.md`.

## Summary by CodeRabbit

- **New Features**
  - Added the `mediaItemChanges` GraphQL feed with revision-based cursors, pagination, aggregate change timestamps, and deletion tombstones.
  - Changes to media items and related library data are now tracked automatically.
  - Added daily processing for date-based availability changes, including startup catch-up.

- **Breaking Changes**
  - Replaced `mediaItems(updatedSince:)` with `mediaItemChanges(first:, after:)`.
  - Existing timestamp-based cursors are no longer supported; clients must begin with a fresh snapshot.

- **Documentation**
  - Updated Library API polling, event, and release guidance.


arsfeld:
Review fixes applied in `1e7ab2fef`. Thanks for the review.

## Accepted and fixed

| Finding | Fix |
|---|---|
| **Install PostgreSQL triggers before backfilling markers** (`migration:83`) | Valid, and the sharpest finding here. `backfill_markers/0` ran before `install_postgres_triggers/0`, so a concurrent `UPDATE`/`DELETE` committing after the backfill snapshot but before `CREATE TRIGGER` acquired its lock left the feed holding a stale marker, or a live marker for a deleted row, with no later revision ever reporting the change. `up/0` now installs functions and triggers before the backfill, in the same transaction: `CREATE TRIGGER` holds a lock that blocks writers until commit, which closes the window entirely. No conflict handling is needed in the backfill because no writer can have produced a marker in between. Verified on a freshly dropped and re-migrated SQLite test database: 12/12 trigger tests pass. |
| **Align `mediaItemChanges` with its revision-feed contract** (`query_types.ex:48`) | Applied verbatim. The field description is now `"Media item changes, oldest revision first. Includes deletions."` and no longer implies a time-bounded window. `priv/graphql/library.graphql` regenerated via `mix library_schema.export`; the SDL test passes. |
| **Em dashes in published prose** (`Strict build` CI failure) | Three occurrences in `docs/using/reference/library-api.md` replaced with parentheses and a colon. |

## Reviewed and declined

| Finding | Reason |
|---|---|
| **Add update triggers for `media_file_episodes`** (`migration:44`) | There is no writer that can UPDATE a join row. Every write to this table is `Repo.insert_all(MediaFileEpisode, rows, on_conflict: :nothing)` (`lib/mydia/library.ex`), both `many_to_many` associations declare `on_replace: :delete` (`lib/mydia/library/media_file.ex:136`, `lib/mydia/media/episode.ex:48`), and a search across `lib/` and `priv/repo/migrations/` finds no `Repo.update` or `update_all` against it. An `AFTER UPDATE OF` trigger would be unreachable code with no test that could fail, and the task brief specified insert/delete for this table. If a future writer ever updates the join, the design spec's "Future fields" contract already requires extending the trigger set at that point. |
| **Use a dedicated gate for `RevisionClockBootstrap`** (`application.ex:326`) | `:start_health_monitors` is `false` in exactly one place, `config/test.exs:84`, so the gate is in practice "is this the test environment", which is what the plan specified. No deployment configuration sets it false, so no install can skip the bootstrap. Introducing a second flag would add configuration surface the plan did not authorize, and the skip exists because the supervisor process has no SQL Sandbox ownership in tests. Worth revisiting if `:start_health_monitors` ever becomes an operator-facing toggle. |

PR #793, opened 2026-09-13
fix(player): allow manual next episode after cancelling up-next auto-play

arsfeld:
## Summary

When a viewer dismisses the **Up Next** auto-play prompt during credits, auto-play correctly stops — but **Next Episode** (transport, keyboard, remote) also stopped working. The viewer had to leave the player and start the next episode manually.

## Root cause

`_cancelAutoPlay()` sets `_autoPlayCancelled = true` for the rest of the file. `_playNextEpisode()` returned early on that flag for **every** call, including explicit manual navigation.

## Fix

- Add `shouldBlockAutoPlayNext()` in `up_next_policy.dart` — only blocks when `fromAutoCountdown && autoPlayCancelled`
- Parameterize `_playNextEpisode({bool fromAutoCountdown = false})`
- Countdown callback passes `fromAutoCountdown: true`; all manual entry points keep the default

## Behavior after fix

| Action | After cancel auto-play |
|--------|------------------------|
| Auto countdown / auto-play | Blocked |
| Manual Next Episode | Works |
| Up Next re-offer | Blocked (unchanged) |

## Test plan

- [x] `flutter test test/presentation/widgets/video_controls/up_next_policy_test.dart`
- [x] `flutter test test/presentation/screens/player/up_next_stop_test.dart`
- [ ] Manual: dismiss Up Next → let episode finish → press Next Episode → next episode starts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Up Next behavior so dismissing the countdown prevents automatic playback while still allowing manually triggered next-episode navigation.
  * Ensured next-episode actions from transport controls, keyboards, and remote controls continue to work as expected after the countdown is dismissed.
  * Prevented dismissed Up Next prompts from triggering unintended episode playback when the countdown expires.

* **Tests**
  * Added coverage for automatic and manual next-episode playback behavior.

### DIFF
--- a/player/lib/presentation/screens/player/player_screen.dart
+++ b/player/lib/presentation/screens/player/player_screen.dart
@@ -3025,7 +3025,9 @@ class _PlayerScreenState extends ConsumerState<PlayerScreen>
 
   void _showUpNextOverlay(UpNextTarget target) {
     _upNextCountdown?.dispose();
-    final countdown = UpNextCountdown(onElapsed: _playNextEpisode);
+    final countdown = UpNextCountdown(
+      onElapsed: () => _playNextEpisode(fromAutoCountdown: true),
+    );
     _upNextCountdown = countdown;
 
     if (!mounted) return;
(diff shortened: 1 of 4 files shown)


arsfeld:
Addressed the CodeRabbit nit on player-level coverage without mounting `PlayerScreen` (same constraint as `player_screen_key_handling_test.dart`).

- Extracted `bindUpNextCountdownElapsed` so the countdown cannot tear-off `_playNextEpisode` and lose `fromAutoCountdown: true`
- Added `player_screen_up_next_next_test.dart`: cancelled countdown does not navigate; a later manual next still does

PR #847, opened 2026-09-18
Sort the library by size, and fix sizes above a terabyte

arsfeld:
Sorts `/movies`, `/tv` and `/sections/:id` by how much disk an item occupies, which is the question you ask when a volume fills up. `LibraryRow` already carried `total_size`, summed in SQL, so the sort is an ordering over data the listing had already loaded.

## Sorting

Two new options in the General optgroup, **Size (Largest)** and **Size (Smallest)**.

Items with nothing on disk sort **last in both directions** rather than leading the ascending list. Sorting them as zero would make "Size (Smallest)" a list of things you do not have. This follows the rule `MydiaWeb.Schema.Resolvers.MediaSort` already states for unknown keys, and the sort stays total, so page boundaries cannot shift between requests during infinite scroll.

## Reading the sizes

The list view's size column now uses `tabular-nums` with the same icon treatment Downloads already uses for this value. Proportional digits made a sorted column look ragged, which rather defeats sorting it.

An item with nothing on disk showed the literal string **"0 B"**: `LibraryRow.total_size` defaults to the integer zero, never nil, so the formatter never reached its nil clause and fell through to the bare-bytes branch. It now shows the same dash the quality column beside it uses.

The header's "Browse and manage your media library" subtitle is replaced by the item count and total bytes on disk for whatever the active search and filters match. Both come from `LibraryListing.page/1`, which already walks every matching row before it paginates, so the figures cost one pass over a list it holds and no extra query. Search for `1080p` and it tells you what your 1080p copies cost.

## Bug fix: sizes above a terabyte

Eight near-identical byte formatters had accumulated in the web layer. Two of them, on the **trash** and **duplicates** pages, had no TB tier and no KB tier, so a 6 TB library read as `6840.5 GB` and a 5 KB file read as `5120 B`. A ninth on the **admin dashboard** had the same gap, so a transcode job over a ~1 TB source read as `1230.5 GB`.

All of them now use `MydiaWeb.Formatters.format_file_size/1`, which already existed with the full five tiers and had no callers.

Three user-visible consequences, all intended:

- Sizes at or above 1 TB render in TB.
- Admin figures move from one decimal to two, so `1.5 GB` may read `1.53 GB`.
- An unknown size renders a dash rather than `N/A` or `unknown`. Only two call sites can actually receive nil, both bare spans over one file's size; every other figure on those pages is a sum that cannot be nil.

## Also removed

`MydiaWeb.LibraryComponents.library_filters/1` was a complete second copy of the search, filter and sort bar that nothing called, while `index.html.heex` rendered its own inline copy. Deleted, so the sort menu has one definition and the two cannot drift.

## Testing

Full suite green: 12,191 tests, 0 failures, with `compile --warnings-as-errors` and `format --check-formatted` clean.

Sort and formatter behaviour is pinned by new tests in `test/mydia/media/library_listing_test.exs`, `test/mydia_web/live/media_live/size_sort_test.exs` and `test/mydia_web/helpers/formatters_test.exs`, including zero-size rows landing last in both directions, tie stability, the filtered total tracking search and filters, and every size tier including TB.

The header stats block was measured in a real browser at 390px: 143px wide, no horizontal overflow.

## Follow-up, not in this PR

Two byte formatters remain in the core layer and still lack a TB tier: `lib/mydia/events/presentation.ex` (trash-sweep event text) and `lib/mydia/indexers/search_result.ex` (release sizes). Neither can import a `MydiaWeb` module without inverting the layering. Fixing them properly means moving the canonical function into the core layer and having the web one delegate.

`mix.exs` still declares `elixir: "~> 1.15"` although every build path is on 1.19. That line is deliberately untouched here because the in-flight Elixir 1.20 work owns it.

## Summary by CodeRabbit

* **New Features**
  * Media library now displays matching item counts and total size.
  * Added sorting by largest or smallest file size.
  * Media list sizes include clearer formatting and an em dash when unavailable.
* **Improvements**
  * File sizes are displayed consistently across downloads, duplicates, trash, system information, and job details.
  * Unknown file sizes now use an em dash instead of “unknown” or “N/A”.
  * Size sorting now provides stable, predictable ordering.
* **Removed**
  * Removed the library’s search, filter, and sort controls component.


arsfeld:
Both findings check out. Taking one, deferring the other.

**Tie determinism (`sort_by_size/2`): fixing here.** You are right, and the case is a bit stronger than the comment suggests: the `unsized` tail was appended with no sort key at all, so on a library with many monitored-but-undownloaded items the entire tail sat in raw scan order. `Media.media_items_query/1` applies no `ORDER BY` and `page/1` rebuilds and re-sorts the whole list on every `load_more`, so on PostgreSQL, where heap order shifts as rows are updated, a row really can be duplicated or skipped across a page boundary. Fixed with explicit secondary keys, zero-size rows still last in both directions, plus a test that pins the tie order rather than inheriting it.

Worth noting for anyone reading later: the other eight `sort/2` clauses in this module share the same latent property, since they all sort a single key over an unordered base query. They are out of scope here but the same fix applies.

**Double-counted multi-episode files (`file_counts/1`): real, deferring to its own PR.** Confirmed: `episode_files` joins `MediaFileEpisode` and sums `mf.size` grouped by `e.media_item_id`, so a file covering two episodes contributes twice. Not taking it here for three reasons:

1. It is pre-existing and this PR does not touch `file_counts/1`. The same wrong total already fed the list view's size column before this change; sorting and the header stat now make it more visible, but they did not cause it.
2. The fix restructures an aggregate that also produces `file_count` and `resolutions`, so it needs its own tests for those, not just for size.
3. It is a data-correctness question worth deciding deliberately: whether a file covering two episodes should count once toward the show's disk usage (yes) has a different answer from how it should count toward per-episode availability, and that deserves more than a quick-win patch on a sorting PR.

Filed as follow-up alongside the two core-layer byte formatters (`lib/mydia/events/presentation.ex` and `lib/mydia/indexers/search_result.ex`) that still lack a TB tier and cannot import the web formatter without inverting the layering.

arsfeld:
Re-raising the `file_counts/1` double-count: still deferring, and your new detail is the reason it should be.

You noted an existing test reaches this case. Confirmed: `test/mydia/media/library_listing_test.exs:86` asserts `row.total_size == 1_400` for one 700-byte file linked to two episodes, and it predates this branch. So the doubling is not an oversight nobody noticed, it is pinned by an assertion someone wrote deliberately. Flipping it is a behaviour decision about what "on disk" means for a file covering two episodes, and it needs to be made with the per-episode availability question in view, not as a quick win on a sorting PR that does not touch `file_counts/1`.

Tracked as follow-up. The tie-determinism finding is fixed in 3bc3a09d5.

PR #894, opened 2026-09-22
feat(relay): list cae1-2 as a second mydia iroh relay

arsfeld:
## Summary

Adds `https://cae1-2.relay.mydia.dev` (the iroh relay on basestar) to the relay list served at `GET /client-config`, next to `cae1-1`. The Worker and the Elixir relay carry the same list, so both change here.

Installs pick this up at their next boot. No server image or player build is needed.

## Verification against the live relay

A throwaway harness (not committed) on the same iroh 1.2 client the server and player use, run from outside basestar:

| Check | Result |
|---|---|
| QUIC address discovery, relay map = cae1-2 only | `udp_v4=true`, public IPv4 returned over UDP 7842 |
| Relay-only echo, IP transports cleared, both ends homed on cae1-2 | 16 MiB round trip, every path `Relay(cae1-2)`, 30 to 35 Mbit/s each way over 3 runs |
| Old client (relay map = cae1-1 only) dials a peer homed on cae1-2 | Connects and echoes 16 MiB |
| Peer homed on cae1-2 dials one homed on cae1-1 | Connects, 22 to 60 Mbit/s over 3 runs |

basestar's `relayserver_accepts_total` went from 0 to 6, matching the connections the harness opened. The third row is the one that matters for rollout: installs still on the old list, or on the compiled-in default, keep reaching servers that choose cae1-2 as their home relay.

Clients choose a home relay by latency across this list plus iroh's public relays, so list order does not pin anyone to cae1-1.

## Deploying

Merging deploys nothing to production. `/client-config` is answered by the Worker (`x-relay-backend: worker`), so the change goes live with a `relay-worker-v*` tag. The Worker's edge cache adds up to an hour on top (`s-maxage=3600`). The `metadata-relay-v0.16.1` tag keeps the Elixir relay's copy in step for the cutover contract diff.

## Test plan

- [x] `relay-worker`: `npx vitest run test/config/client_config.test.ts` (5 passed), `npm run typecheck`
- [x] `metadata-relay`: `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix test test/metadata_relay/client_config_test.exs` (4 passed)
- [ ] After the tags: `curl https://relay.mydia.dev/client-config` lists both relays


arsfeld:
Superseded by #895, which makes the relay list configuration instead of code. cae1-2 now ships as a `client-config:relays` KV write, with no code change and no tag. See relay-worker/README.md, "Changing the relay list".

PR #900, opened 2026-09-22
Player remote logs

arsfeld:
Players can now send their logs to the metadata relay, so a player problem that
is not a crash can be investigated without getting at the device.

**Player**

- Every `debugPrint`, and every Rust p2p `tracing` event, is captured into
  rolling NDJSON files in the app support directory (about 10 MB), with secrets
  removed as they are written: tokens, passwords, auth headers and
  credential-bearing query params, including quoted values with spaces in them.
- Settings' "Connection details" row is now "Diagnostics". Its screen has a
  "Share with developers" section that replaces the crash reporting toggle:
  off, crashes only, or crashes and logs for 24 hours, 7 days or until turned
  off. The old crash consent migrates on first launch and its key is deleted.
- "Send logs now" uploads the recent log once and shows a code like
  `LOG-7K2QX9` to quote in an issue.
- Nothing timestamped after the sharing window is ever uploaded, including when
  sharing is turned off while an upload is in flight.

**Relay**

- `POST /player-logs` takes gzipped NDJSON batches: 1 MB compressed, 8 MB
  decompressed with incremental decompression, rate limited per address and per
  device, 50 MB a day per device, and the quota is only charged once a batch is
  actually stored.
- Batches are gzipped files beside the SQLite database, which holds only the
  index. Stream batches are kept 14 days, reports 90, under a 2 GiB cap that
  evicts the oldest stream batches first. Every sweep step is bounded per run.
- `GET /logs/raw` (dashboard basic auth) prints a device's or a report's lines
  as plain text, journalctl style, streamed one stored file at a time so memory
  stays flat. `/logs` lists devices and reports.

**Tooling**

- `./dev relay logs <device|LOG-code> [--since 1h] [--level warn] [--tag X] [--grep text]`.

Verified end to end against a local relay: a report upload returning a code, a
stream batch, reading both back with filters, and the dashboard refusing
unauthenticated requests. Relay suite 395, player suite 4357, Rust crate 14,
analyzer 0 errors.

Rollout: no new secrets. Check free space on the relay's `/data` volume before
tagging, and set `PLAYER_LOGS_MAX_BYTES` if 2 GiB does not fit. The relay
version is bumped to 0.18.0 because 0.17.0 is already tagged.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in Diagnostics sharing for crashes and player logs, with 24-hour, 7-day, and indefinite options, plus “Send logs now” with report codes.
  * Added player-log upload and retrieval, including filtering, retention, authenticated dashboards, and a `dev relay logs` command.
  * Player logs can upload continuously while sharing is active and flush when the app backgrounds; sensitive values are redacted before storage.
* **Reliability**
  * Added per-device and per-address upload limits, with safeguards against exceeding quotas during concurrent uploads.
* **Documentation**
  * Documented player-log configuration, limits, retention, and retrieval.


arsfeld:
@coderabbitai review

